PR #265, opened 2014-12-12
Allow several references from a property to the same node

### LINKED ISSUE #242: ReferenceMany association with several references to the same node
originally created here https://github.com/doctrine/DoctrinePHPCRBundle/issues/156

this looks like an edge case that we just did not think about. i just tested with JavaDavexClient and indeed it works there to have two times the same reference. so this is a bug we have.

now to fix it, i am a bit unsure. we have a Session::getNodesByIdentifier that says that the keys are the uuid. so there, duplicates are eliminated. the ObjectManager::getNodesByIdentifier makes the uuids unique, which at least for fetching from the backend makes sense.

maybe we should have a helper method in the getNode method for both weak and hard references that maps over $value and looks up the ids in $nodes, throwing the exception as we have now when a node is not found, and puts the nodes into $results without using an explicit key. that way, the ordering is guaranteed to be correct and we can have duplicates again.

dbu:
fix #242

tested by https://github.com/phpcr/phpcr-api-tests/pull/145

### DIFF
--- a/src/Jackalope/ObjectManager.php
+++ b/src/Jackalope/ObjectManager.php
@@ -603,7 +603,7 @@ class ObjectManager
      */
     public function getNodesByIdentifier($identifiers, $class = 'Node')
     {
-        $nodes = $fetchIdentifiers = array();
+        $nodes = $fetchPaths = array();
 
         foreach ($identifiers as $uuid) {
             if (!empty($this->objectsByUuid[$uuid])
@@ -613,6 +613,7 @@ class ObjectManager
                 $nodes[$uuid] = $this->objectsByPath[$class][$this->objectsByUuid[$uuid]];
             } else {
                 $fetchPaths[$uuid] = $uuid;
+                $nodes[$uuid] = $uuid; // keep position
             }
         }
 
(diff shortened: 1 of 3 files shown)


lsmith77:
is there a reason why this PR didn't go into 1.1?

dbu:
i guess its an edge case. its somewhat bugfix, somewhat new feature. but as we should be able to release 1.2 very soon i think we should not bother with backporting this.

lsmith77:
it seems urgent for the sulu guys https://github.com/jackalope/jackalope/pull/276

dantleech:
I'm pretty sure this caused a BC break as the nodes now no longer have the UUIDs as keys. Is this a requirement for this PR, or can we reintroduce the UUID as a key in the returned array?

dbu:
this was an undocumented thing. and we can't have both. if we use the uuid as key, there can be no duplicates as each key is unique, adding the thing again will overwrite the value, not add a second one.

dantleech:
I remember now.. ok I guess we will have to add a workaround until the next version is released. /cc @wachterjohannes

wachterjohannes:
@dantleech ok then we should wait i have created a "hack" for the client project to avoid errors there and hope that this is released as soon as possible to remove all hacks out of sulu! (=

PR #272, opened 2015-01-26
Use class cache in factory

EmmanuelVella:
This avoid a lot of ``class_exists`` calls, saving some precious milliseconds !

### DIFF
--- a/src/Jackalope/Factory.php
+++ b/src/Jackalope/Factory.php
@@ -16,6 +16,11 @@ class Factory implements FactoryInterface
     /**
      * @var array
      */
+    protected $classCache = array();
+    
+    /**
+     * @var array
+     */
     protected $reflectionCache = array();
 
     /**
@@ -23,10 +28,18 @@ class Factory implements FactoryInterface
      */
     public function get($name, array $params = array())
     {
-        if (class_exists('Jackalope\\' . $name)) {
-            $name = 'Jackalope\\' . $name;
-        } elseif (! class_exists($name)) {
-            throw new InvalidArgumentException("Neither class Jackalope\\$name nor class $name found. Please check your autoloader and the spelling of $name");
+        if (isset($this->classCache[$name])) {
+            $name = $this->classCache[$name];
+        } else {
+            $originalName = $name;
+
+            if (class_exists('Jackalope\\' . $name)) {
+                $name = 'Jackalope\\' . $name;
+            } elseif (! class_exists($name)) {
+                throw new InvalidArgumentException("Neither class Jackalope\\$name nor class $name found. Please check your autoloader and the spelling of $name");
+            }
+            
+            $this->classCache[$originalName] = $name;
         }
 
         if (0 === strpos($name, 'Jackalope\\')) {


lsmith77:
can you give some information about the relative savings this gives?

EmmanuelVella:
Sure ! I made a quick benchmark on my local machine, on a symfony2 prod environment :

 | Before | After
---- | ------------ | -------------
Calls | 5332 | 81
Total time | 485 ms | 460 ms

lsmith77:
while you are at it .. I always wondered if https://github.com/jackalope/jackalope/pull/272/files#diff-710f3ac2ecf6432c840eacf5dba8209eR60 is causing a noticeable amount of overhead. ie.if it would make sense to hardcode a few cases (like 1, 2, 3, 4 constructor parameters).

EmmanuelVella:
Here is a benchmark with 5 000 (realist) and 10 000 000 iterations : 

      | 5 000 | 10 000 000
---- | ------------ | -------------
Without reflection cache | ~14 ms | 25672 ms
With reflection cache | ~6 ms | 20405 ms

EmmanuelVella:
I just did another benchmark for the class cache I added in this PR : 

      | 5 000 | 10 000 000
---- | ------------ | -------------
Without class cache | ~24 ms | 52312 ms
With class cache | ~6 ms | 20405 ms

lsmith77:
@dbu fine for me to merge this .. 

@EmmanuelVella did you also try to benchmark if hardcoding different permutations of constructor parameter counts performs faster than using Reflection for all cases?

lsmith77:
ie something like this (I did a quick check and 6 seems to be the maximum number of constructor params the use):
```
diff --git a/src/Jackalope/Factory.php b/src/Jackalope/Factory.php
index d9b0a0c..89ed384 100644
--- a/src/Jackalope/Factory.php
+++ b/src/Jackalope/Factory.php
@@ -28,12 +28,25 @@ class Factory implements FactoryInterface
             array_unshift($params, $this);
         }
 
-        if (! count($params)) {
-            return new $name;
-        }
-
-        $class = new ReflectionClass($name);
+        switch (count($params)) {
+            case 0:
+                return new $name();
+            case 1:
+                return new $name($params[0]);
+            case 2:
+                return new $name($params[0], $params[1]);
+            case 3:
+                return new $name($params[0], $params[1], $params[2]);
+            case 4:
+                return new $name($params[0], $params[1], $params[2], $params[3]);
+            case 5:
+                return new $name($params[0], $params[1], $params[2], $params[3], $params[4]);
+            case 6:
+                return new $name($params[0], $params[1], $params[2], $params[3], $params[4], $params[5]);
+            default:
+                $class = new ReflectionClass($name);
 
-        return $class->newInstanceArgs($params);
+                return $class->newInstanceArgs($params);
+        }
     }
 }
```

EmmanuelVella:
Here are the results : 

      | 5 000 | 10 000 000
---- | ------------ | -------------
With ReflectionClass  | ~12 ms | 23588 ms
With switch| ~10 ms | 16981 ms

IMHO the ``switch`` makes the code more complicated and do not provide a noticeable performance improvement (in a normal use case).

dbu:
thanks a lot! 
we started thinking about jackalope 2.0 - i think the factory is one place to think again. i added some ideas to https://github.com/jackalope/jackalope/wiki/2.0#factory

EmmanuelVella:
:+1: thanks for merging this

PR #273, opened 2015-01-29
use LIBXML_PARSEHUGE option to be able to import data with large binaries

uwej711:
I had an issue with an export that could not be imported. There was a huge binary included on which the XMLReader obviously choked. Works with that extra option.

### DIFF
--- a/src/Jackalope/ImportExport/ImportExport.php
+++ b/src/Jackalope/ImportExport/ImportExport.php
@@ -95,7 +95,7 @@ class ImportExport implements ImportUUIDBehaviorInterface
         }
 
         $xml = new FilteredXMLReader;
-        $xml->open($uri);
+        $xml->open($uri, null, LIBXML_PARSEHUGE);
         if (libxml_get_errors()) {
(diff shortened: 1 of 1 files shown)


src/Jackalope/ImportExport/ImportExport.php
```diff
@@ -95,7 +95,7 @@ public static function importXML(NodeInterface $parentNode, NamespaceRegistryInt
         }
 
         $xml = new FilteredXMLReader;
-        $xml->open($uri);
+        $xml->open($uri, null, LIBXML_PARSEHUGE);
```
dbu: according to http://php.net/manual/en/libxml.constants.php this exists since php 5.3.2 which is fine, but only with libxml 2.7+ . can you please add a version check and only add this if the libxml version is 2.7+?

uwej711: OK, just had to really search a while how to check that version ...

dbu:
great, thanks a lot for this fix!

PR #274, opened 2015-02-10
[WIP] Acl

lsmith77:


src/Jackalope/Security/AccessControlEntry.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @var PrivilegeInterface[]
+     */
+    private $privileges;
+
+    public function __construct(PrincipalInterface $principal, array $privileges)
+    {
+        $this->principal = $principal;
+        // TODO: validate privileges
```
lsmith77: probably not needed, since the privilege instances will already be "validated"

src/Jackalope/Security/AccessControlManager.php
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            foreach ($privilege->getAggregatePrivileges() as $childPrivilege) {
+                if ($privilegeName === $childPrivilege->getName()) {
+                    return $childPrivilege;
+                }
+            }
+        }
+
+        throw new AccessControlException($privilegeName);
```
lsmith77: Improve exception message

src/Jackalope/Security/AccessControlManager.php
```diff
@@ -0,0 +1,140 @@
...
+    {
+        return $this->om->getPolicies($absPath);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getEffectivePolicies($absPath)
+    {
+        throw new RepositoryException('This method can not properly be implemented. Use getPolicies for this path');
```
lsmith77: currently not implemented for JCR remoting in Jackrabbit

src/Jackalope/Security/Privilege.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return $this->name;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function isAbstract()
+    {
+        return false;
```
lsmith77: we need to make this dynamic, since it is possible to set a privilege as abstract in Jackrabbit

src/Jackalope/Session.php
```diff
@@ -663,7 +670,15 @@ public function isLive()
      */
     public function getAccessControlManager()
     {
-        throw new UnsupportedRepositoryOperationException();
+        if (!$this->getTransport() instanceof AccessControlInterface) {
+            //throw new UnsupportedRepositoryOperationException();
```
lsmith77: re-enable this

src/Jackalope/Security/AccessControlEntry.php
dbu: what i meant is validate that its actually PrivilegeInterface instances, in case somebody tries to set them with strings or something like that.

src/Jackalope/ObjectManager.php
```diff
@@ -940,6 +946,10 @@ protected function executeBatch($type, $operations)
             case Operation::REMOVE_PROPERTY:
                 $this->transport->deleteProperties($operations);
                 break;
+            case Operation::SET_POLICY:
+                $this->transport->setPolicy($operations);
+                break;
+            //TODO: operation for acl changes
```
dbu: afaik there is no different operation, the acl is always entirely overwritten. right @Alfusainey ?

src/Jackalope/Session.php
dbu: this was because i missed the LoggingClient, can just be undone.

src/Jackalope/Security/Privilege.php
dbu: afaik no, phpcr/jcr seems to have no way to make something abstract. ah, or its an implementation specific extension? then we would need to know how that fact is communicated by the server , and probably add a flag in the constructor.

src/Jackalope/Security/AccessControlManager.php
```diff
@@ -0,0 +1,140 @@
...
+
+        $this->om->setPolicy($absPath, $policy);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function removePolicy($absPath, AccessControlPolicyInterface $policy)
+    {
+        // TODO: track this so its saved
```
dbu: need an operation for this

src/Jackalope/Security/AccessControlManager.php
dbu: yep, and angela thinks it makes semantically no sense, as you might not have the right to read all policies that might affect you, and there is no defined behaviour for that case. the things that are releavant are: 
* what policies are applicable at a specific path?
* do i have the permission to do something? this is SessionInterface::hasPermission

src/Jackalope/ObjectManager.php
dbu: but there is one for REMOVE_POLICY

PR #276, opened 2015-02-13
Fix sorting for 1.1 release

danrot:


tests/Jackalope/PropertyTest.php
```diff
@@ -2,8 +2,67 @@
...
+                'Internal Error: Could not find a referenced node. If the referencing node is a frozen version, this can happen, otherwise it would be a bug.',
+            ),
+            array(
+                array('1234-1234', '4321-4321'),
+                array('1234-1234', '4321-4321'),
+            ),
+            array(
+                array('1234-1234', '4321-4321'),
+                array('4321-4321', '1234-1234'),
+                'Internal Error: Could not find a referenced node. If the referencing node is a frozen version, this can happen, otherwise it would be a bug.',
```
danrot: @lsmith77 I had to manually add this line, because this test was succeeding before. But this behavior is exactly what is failing for us, I think this should result in this error.

danrot: Found the reason, it is that on master the check at https://github.com/jackalope/jackalope/blob/1.1/src/Jackalope/Property.php#L450 has been removed. Actually that's exactly what makes our application fail, but we really need the correct order.

dbu:
@danrot are you backporting #265 to 1.1? as we are about to release jackalope 1.2 (i would say by end of february we should reach stable), could we instead just add your test to the master branch?

if it is important to get this fixed in 1.1  we can also merge this into 1.1, i don't mind. but then i am confused by your comment - you only change tests and say that they had been removed in master...

danrot:
Yes, I would need the change in #265 to correctly implement sorting in Sulu. It would be nice to have this change in 1.1

I only had to adjust the tests, because the commit I have cherry-picked caused the tests to fail. This is why I change the tests. I basically added the same lines as in master, but they haven't been in a single commit, so I couldn't cherry-pick them.

dbu:
okay. i guess i will merge then, and solve an "interesting" merge conflict to master :-/

let me just test things locally one more time to be sure

danrot:
I just hope that isn't too much work then...

dbu:
i will manage :-)

i just realized that maintaining version 1.1 of jackalope already sucks a bit because the phpcr-api-tests evolve continually and basically every test we add to verify some fix means that older versions of jackalope will fail the test suite. i created https://github.com/phpcr/phpcr-api-tests/issues/154 to discuss that problem.

lsmith77:
tricky .. I guess we need to then disable missing stuff

lsmith77:
@dbu: will you also tag a release?

tests/Jackalope/PropertyTest.php
dbu: hm, the mystery deepends. when backporting this to master branch, i had to remove this line again as the order no longer seems to matter. #277 keeps this message removed and tests succeed.

that line 450 would make no sense as the getNodesByIdentifiers eliminates duplicate ids. Property::getReferencedNodes should restore the corect order however. do things work for you now with the latest 1.1 branch?

dbu: ah, now i am less confused. you seem to have backported the fix for weakreferences but not for hard references. see the lines for case: REFERENCE, which still have the old behaviour.

PR #277, opened 2015-02-13
update master from 1.1

dbu:
merging changes from 1.1 branch to master.

### DIFF
--- a/tests/Jackalope/PropertyTest.php
+++ b/tests/Jackalope/PropertyTest.php
@@ -2,8 +2,6 @@
 
 namespace Jackalope;
 
-use Jackalope\Factory;
-use Jackalope\TestCase;
 use PHPCR\PropertyType;
 
 class PropertyTest extends TestCase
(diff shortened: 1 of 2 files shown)


tests/Jackalope/PropertyTest.php
```diff
@@ -14,12 +12,12 @@ public function provideGetNode()
             array(
                 array('1234-1234', '4321-4321'),
                 array(),
-                'Could not find one or more referenced nodes: "1234-1234", "4321-4321"',
+                'Internal Error: Could not find one or more referenced nodes: "1234-1234", "4321-4321". If the referencing node is a frozen version, this can happen, otherwise it would be a bug.',
```
dbu: actually the previous test was good enough, now its extremly specific but whatever, not a problem.

danrot: Really? The tests were failing for me, because it seems like the exception message has changed.

dbu: compared to 1.1 the message changed indeed. but in master we only had the relevant  fragment of the message, which is enough for phpunit.

PR #278, opened 2015-03-17
Use class cache in factory

EmmanuelVella:
I made my https://github.com/jackalope/jackalope/pull/272 PR on the master branch, but I think it would be nice to have it on the ``1.1`` branch to !

Is it be possible ?

### DIFF
--- a/src/Jackalope/Factory.php
+++ b/src/Jackalope/Factory.php
@@ -14,14 +14,32 @@ use ReflectionClass;
 class Factory implements FactoryInterface
 {
     /**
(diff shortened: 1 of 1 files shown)


dbu:
i am -1 on porting this to 1.1. rather we should release 1.2 very soon. afaik there is only 1 blocker left: https://github.com/jackalope/jackalope-doctrine-dbal/pull/223

EmmanuelVella:
Ok no problem, I though the next release would be 2.0. Thank you for your feeback !

PR #279, opened 2015-03-21
fix issues with the node processor

dbu:
Fix some issues found while working on https://github.com/jackalope/jackalope-doctrine-dbal/pull/223


src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -54,18 +58,42 @@ class NodeProcessor
...
      * Regex to assert the a string is a valid PATH
+     *
+     * TODO: better regexp for this validation.
+     *
+     * We had /^((\/|..)?[-a-zA-Z0-9:_]+)+$/ which is too strict.
+     * We need to allow many more characters in the name, and we need to allow
+     * relative paths.
      */
-    const VALIDATE_PATH = '/^((\/|..)?[-a-zA-Z0-9:_]+)+$/';
+    const VALIDATE_PATH = '(((/|..)?[-a-zA-Z0-9:_]+)+)';
```
dbu: this expression is rather pointless, requiring the path to contain once anywhere something of the range `[-a-zA-Z0-9:_]`. but the intention is quite wrong, and the expression above breaks many things. should we validate at all? how can we validate this in a sane way? it can be any path-valid character which is almost everything, and PathHelper::assertValidAbsolutePath does not work as the path may be relative in this case...

ideas what to do? drop the check? create PathHelper::assertValidRelativPath?

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -81,9 +109,11 @@ public function __construct(
     /**
      * Process the given node.
      *
+     * TODO: Either we should have this method return the additional operations, or call getAdditionalOperations
```
dbu: why does this method not return the additional operations directly? and it seems in the integration, we missed to handle this.

tests/Jackalope/NodeType/NodeProcessorTest.php
```diff
@@ -306,25 +311,27 @@ public function providePropertyValidation()
...
                 array(
                     'getType' => PropertyType::PATH,
                     'isMultiple' => false,
                     'getValue' => '  pathto££333+_123£³[]/something[&&"£$]',
                     'getPath' => '/path/to',
-                ), 
+                ),
                 'Value "  pathto££333+_123£³[]/something[&&"£$]" for PATH property at "/path/to" is invalid',
             ),
+            */
```
dbu: commented this out for now, as i made the PATH validation too lax again

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -54,18 +57,34 @@ class NodeProcessor
...
     /**
      * Regex to assert the a string is a valid PATH
+     *
+     * TODO: better regexp for this validation.
+     *
+     * We had /^((\/|..)?[-a-zA-Z0-9:_]+)+$/ which is too strict.
+     * We need to allow many more characters in the name, and we need to allow
+     * relative paths.
+     */
+    const VALIDATE_PATH = '(((/|..)?[-a-zA-Z0-9:_]+)+)';
```
dbu: this expression is rather pointless, requiring the path to contain once anywhere something of the range `[-a-zA-Z0-9:_]`. but the intention is quite wrong, and the expression we have here in master breaks many things. how can we validate this in a sane way? should we just skip validation or fallback to something extremly basic? PATH may be relative, and may contain any path-valid character which is almost everything, and PathHelper::assertValidAbsolutePath does not work as it requires the path to be absolute.

ideas what to do? drop the check? create PathHelper::assertValidPath that does not enforce an absolute path?

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -73,45 +92,51 @@ public function __construct(
...
         $this->namespaces = $namespaces;
     }
 
     /**
-     * Process the given node.
+     * Process the given node and return eventual extra operations determined from the node.
      *
      * @param NodeInterface $node
+     *
+     * @return AddNodeOperation[] Additional operations that the client must execute for autocreated nodes.
```
dbu: refactored to return the operations directly in process, instead of member variables.

dbu:
in the interest of moving forward i propose that i do one of these:

* use PathHelper::normalizePath relative to the current node, to validate the path at least a little bit
* stop validating at all (as it was until the last version)

what do you prefer?

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -262,12 +279,11 @@ private function assertValidProperty($property)
...
                 foreach ($values as $value) {
-                    if (!preg_match(self::VALIDATE_PATH, $value)) {
-                        throw new ValueFormatException(sprintf(
-                            'Value "%s" for PATH property at "%s" is invalid. Segments are separated by / and allowed chars are -a-zA-Z0-9:_',
-                            $value,
-                            $property->getPath()
-                        ));
+                    try {
+                        // making the path absolute also validates the result to be a valid path
+                        PathHelper::absolutizePath($value, $property->getNode()->getPath());
```
dbu: what do you think of this solution? this will be very permissive, but consistent with other path validations.

dbu:
okay, i think this is ready. any last inputs? good to merge?

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -262,12 +279,11 @@ private function assertValidProperty($property)
...
                 foreach ($values as $value) {
-                    if (!preg_match(self::VALIDATE_PATH, $value)) {
-                        throw new ValueFormatException(sprintf(
-                            'Value "%s" for PATH property at "%s" is invalid. Segments are separated by / and allowed chars are -a-zA-Z0-9:_',
-                            $value,
-                            $property->getPath()
-                        ));
+                    try {
+                        // making the path absolute also validates the result to be a valid path
+                        PathHelper::absolutizePath($value, $property->getPath());
```
dbu: what do you think of this solution? this will be very permissive, but consistent with other path validations. we interpret the path relative to the property instead of the node, but that does not matter. we could even just absolutize against `/`.

dbu:
@dantleech / @lsmith77 can one of you please review this and merge? i guess then we can tag beta1 or RC1 and i updated https://github.com/jackalope/jackalope-doctrine-dbal/pull/223 to rely on that.

lsmith77:
I do not know the details of the previous change to really comment. overall I like having less state being used. so +1 from me.

src/Jackalope/NodeType/NodeProcessor.php
dantleech: Does this change the existing behavior? I think I stole the RegEx from the client.

dbu: yeah, the one in the client was pointless. it had unescaped `.` in it so
did not validate anything. what i do now is checking that this is
potentially a valid path, although implementation specific restrictions
are not handled this way but that is an issue throughout the code, as
PathHelper has static methods and thus can not be extended to add
restrictions.
so imo its better than before and we can solve the issue of proper path
validation at a different day.

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -73,45 +82,51 @@ public function __construct(
...
             /* @var $nodeType NodeTypeDefinitionInterface */
-            $this->processNodeWithType($node, $nodeType);
+            $additionalOperations = array_merge(
+                $additionalOperations,
+                $this->processNodeWithType($node, $nodeType)
+            );
         }
-    }
 
-    public function getAdditionalOperations()
```
dantleech: +1 I think. Only used in the tests and we now aggregate the additional operations locally in `process` from what I see.

dantleech:
Looks good to me

PR #280, opened 2015-04-04
[WIP] Versioning

danrot:
This PR tries to implement versioning as specified in http://www.day.com/specs/jcr/2.0/15_Versioning.html

Since all the versioning information is also saved in the tree as nodes and properties this can be implemented transport agnostic.

### DIFF
--- a/src/Jackalope/NodeType/NodeProcessor.php
+++ b/src/Jackalope/NodeType/NodeProcessor.php
@@ -8,6 +8,7 @@ use PHPCR\NodeType\NodeTypeDefinitionInterface;
 use PHPCR\NodeType\PropertyDefinitionInterface;
 use PHPCR\PropertyInterface;
 use PHPCR\PropertyType;
+use Jackalope\Version\VersionHandler;
 use PHPCR\RepositoryException;
 use PHPCR\Util\PathHelper;
 use PHPCR\ValueFormatException;
@@ -161,12 +162,14 @@ $/xi";
             }
 
             if (!$node->hasProperty($propertyDef->getName())) {
-                if ($propertyDef->isMandatory() && !$propertyDef->isAutoCreated()) {
+                if ($propertyDef->isMandatory() && !$propertyDef->isAutoCreated() &&
+                    $nodeTypeDefinition->getName() !== VersionHandler::MIX_VERSIONABLE
+                ) {
                     throw new RepositoryException(sprintf(
                         'Property "%s" is mandatory, but is not present while saving "%s" at "%s"',
-                         $propertyDef->getName(),
-                         $nodeTypeDefinition->getName(),
-                         $node->getPath()
+                        $propertyDef->getName(),
+                        $nodeTypeDefinition->getName(),
+                        $node->getPath()
                     ));
                 }
 
@@ -193,7 +196,7 @@ $/xi";
                                 $value = $defaultValues;
                             } elseif (isset($defaultValues[0])) {
                                 $value = $defaultValues[0];
-                            } else {
+                            } elseif ($nodeTypeDefinition->getName() !== VersionHandler::MIX_VERSIONABLE) {
                                 // When implementing versionable or activity, we need to handle more properties explicitly
                                 throw new RepositoryException(sprintf(
                                     'No default value for autocreated property "%s" at "%s"',
@@ -212,7 +215,7 @@ $/xi";
             } elseif ($propertyDef->isAutoCreated()) {
                 $prop = $node->getProperty($propertyDef->getName());
                 if (!$prop->isModified() && !$prop->isNew()) {
-                    switch($propertyDef->getName()) {
+                    switch ($propertyDef->getName()) {
                         case 'jcr:lastModified':
                             if ($this->autoLastModified) {
                                 $prop->setValue(new \DateTime());
@@ -269,7 +272,7 @@ $/xi";
                                 $property->getPath(),
                                 $prefix
                             ));
-                            }
+                        }
                     }
                 }
                 break;
@@ -304,7 +307,7 @@ $/xi";
                 break;
             case PropertyType::DECIMAL:
             case PropertyType::STRING:
-                $values = (array) $property->getValue();
+                $values = (array)$property->getValue();
                 foreach ($values as $value) {
                     if (0 !== preg_match(self::VALIDATE_STRING, $value)) {
                         throw new ValueFormatException(sprintf(
--- a/src/Jackalope/NodeType/NodeProcessor.php
+++ b/src/Jackalope/NodeType/NodeProcessor.php
@@ -2,6 +2,7 @@
 
 namespace Jackalope\NodeType;
 
+use Doctrine\Common\Version;
 use PHPCR\NodeInterface;
 use PHPCR\NodeType\NodeDefinitionInterface;
 use PHPCR\NodeType\NodeTypeDefinitionInterface;
@@ -73,6 +74,11 @@ $/xi";
     private $namespaces = array();
 
     /**
+     * @var VersionHandler
+     */
+    private $versionHandler;
+
+    /**
      * @param string  $userId           ID of the connected user
      * @param array   $namespaces       List of namespaces in the current session. Keys are prefix, values are URI.
      * @param boolean $autoLastModified Whether the last modified property should be updated automatically
@@ -80,12 +86,14 @@ $/xi";
     public function __construct(
         $userId,
         $namespaces = array(),
-        $autoLastModified = true
+        $autoLastModified = true,
+        VersionHandler $versionHandler = null
     )
     {
         $this->userId = (string) $userId;
         $this->autoLastModified = $autoLastModified;
         $this->namespaces = $namespaces;
+        $this->versionHandler = $versionHandler;
     }
 
     /**
@@ -128,6 +136,10 @@ $/xi";
     {
         $additionalOperations = array();
 
+        if ($nodeTypeDefinition->isNodeType(VersionHandler::MIX_SIMPLE_VERSIONABLE)) {
+            $additionalOperations = $this->versionHandler->addVersionProperties($node);
+        }
+
         foreach ($nodeTypeDefinition->getDeclaredChildNodeDefinitions() as $childDef) {
             /* @var $childDef NodeDefinitionInterface */
             if (!$node->hasNode($childDef->getName())) {
@@ -162,8 +174,9 @@ $/xi";
             }
 
             if (!$node->hasProperty($propertyDef->getName())) {
-                if ($propertyDef->isMandatory() && !$propertyDef->isAutoCreated() &&
-                    $nodeTypeDefinition->getName() !== VersionHandler::MIX_VERSIONABLE
+                if ($propertyDef->isMandatory()
+                    && !$propertyDef->isAutoCreated()
+                    && $nodeTypeDefinition->getName() !== VersionHandler::MIX_VERSIONABLE
                 ) {
                     throw new RepositoryException(sprintf(
                         'Property "%s" is mandatory, but is not present while saving "%s" at "%s"',
--- a/src/Jackalope/ObjectManager.php
+++ b/src/Jackalope/ObjectManager.php
@@ -5,6 +5,8 @@ use ArrayIterator;
 use InvalidArgumentException;
 
 use Jackalope\Transport\NodeTypeFilterInterface;
+use Jackalope\Version\VersionHandler;
+use PHPCR\NodeType\NodeTypeInterface;
 use PHPCR\SessionInterface;
 use PHPCR\NodeInterface;
 use PHPCR\PropertyInterface;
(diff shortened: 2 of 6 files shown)


src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -178,13 +185,15 @@ private function processNodeWithType(Node $node, NodeType $nodeTypeDefinition)
                                 $value = $defaultValues;
                             } elseif (isset($defaultValues[0])) {
                                 $value = $defaultValues[0];
-                            } else {
+                            } elseif ($nodeTypeDefinition->getName() !== VersionHandler::MIX_VERSIONABLE) {
```
danrot: I have stolen this implementation detail from Jackrabbit: http://svn.apache.org/repos/asf/jackrabbit/trunk/jackrabbit-core/src/main/java/org/apache/jackrabbit/core/ItemSaveOperation.java

The versioning information is added at a later point, since consistency checks make it hard (or even impossible?) to add them before or during the check done by the `NodeProcessor`.

src/Jackalope/ObjectManager.php
```diff
@@ -829,6 +831,13 @@ public function save()
                         if ($node->needsChildReordering()) {
                             $this->transport->reorderChildren($node);
                         }
+                        // add versioning properties
+                        foreach($node->getMixinNodeTypes() as $mixinType) {
```
danrot: Wouldn't be necessary if the `NodeInterface` contains a `hasMixin`-method. Can we add this to the Interface, or are we strictly following the specification here?

src/Jackalope/Version/VersionHandler.php
```diff
@@ -0,0 +1,66 @@
...
+            $versionStorageNode->addNode($node->getIdentifier(), 'nt:versionHistory');
+        }
+        $versionHistory = $versionStorageNode->getNode($node->getIdentifier());
+        $versionHistory->setProperty('jcr:versionableUuid', $node->getIdentifier());
+
+        // TODO Set jcr:copiedFrom if needed
+
+        $versionHistory->addNode('jcr:versionLabels', 'nt:versionLabels');
+        $rootVersion = $versionHistory->addNode('jcr:rootVersion', 'nt:version');
+        $reflection->invoke($transport, $rootVersion->getPath(), $rootVersion->getProperties());
```
danrot: That's horrible, but I couldn't find a way around using reflections...

The problem here is that the Nodes have to be persisted before they can be referenced. That's why I call here (and a few lines later) the protected `storeNode` method via a reflection. That's only necessary because there is no way to save a single node. If `$session->save()` is called the system ends up in an inconsistent state (because some nodes are being created twice) and a rollback is executed.

How should we handle this? /cc @dbu

src/Jackalope/Version/VersionHandler.php
```diff
@@ -0,0 +1,66 @@
...
+        $session = $node->getSession();
+        $rootNode = $session->getRootNode();
+
+        $node->setProperty('jcr:isCheckedOut', true);
+
+        if (!$rootNode->hasNode('jcr:system/jcr:versionStorage')) {
+            if (!$rootNode->hasNode('jcr:system')) {
+                $rootNode->addNode('jcr:system');
+            }
+            $rootNode->addNode('jcr:system/jcr:versionStorage');
```
danrot: I am not sure if it would be better to bootstrap this while initializing the system /cc @dbu

src/Jackalope/ObjectManager.php
dantleech: You can use isNodeType() I think.

src/Jackalope/Version/VersionHandler.php
dantleech: In Jackalope FS I had this logic in a [separate class](https://github.com/jackalope/jackalope-fs/blob/master/src/Transport/Fs/Filesystem/Storage/NodeWriter.php#L41). Which is might be the approach to take in the future. But for now I see now harm in making the method public ?

src/Jackalope/Version/VersionHandler.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Jackalope\Version;
+
+use Jackalope\Node;
+use Jackalope\Transport\TransportInterface;
+use PHPCR\UnsupportedRepositoryOperationException;
+
+class VersionHandler
+{
+    const MIX_VERSIONABLE = 'mix:versionable';
+    const MIX_SIMPLE_VERSIONABLE = 'mix:simpleVersionable';
```
dantleech: We should probably put all the specification node type names in a single constant class.

src/Jackalope/Version/VersionHandler.php
dantleech: If we did, it could be when the (workspace is initiated)[https://github.com/jackalope/jackalope-doctrine-dbal/blob/master/src/Jackalope/Transport/DoctrineDBAL/Client.php].

src/Jackalope/Version/VersionHandler.php
dantleech: Same for the property names

src/Jackalope/NodeType/NodeProcessor.php
dantleech: I wonder if we could refactor this, something like:

````php
if ($propertyDef->isAutoCreated()) {
    $value = null;
    if ($nodeType != 'mix:versionable') {
        $value = $this->getDefaultValue($propertyDef);
    }
}
````

Otherwise is `$value` undefined if the node is MIX_VERSIONABLE?

Comment (danrot) [inline src/Jackalope/NodeType/NodeProcessor.php]: The thing is that these values are not auto created...

File: src/Jackalope/Version/VersionHandler.php (hunk 3 above)
Comment (danrot) [inline src/Jackalope/Version/VersionHandler.php]: Well, it's something which might easily break BC in the future :-)

File: src/Jackalope/Version/VersionHandler.php (hunk 5 above)
Comment (dbu) [inline src/Jackalope/Version/VersionHandler.php]: NodeTypeInterface has constants for these, but not using the namespace alias but the full namespace: https://github.com/phpcr/phpcr/blob/master/src/PHPCR/NodeType/NodeTypeInterface.php#L257

File: src/Jackalope/Version/VersionHandler.php [hunk 6]
```diff
@@ -0,0 +1,66 @@
...
+     * @var TransportInterface
+     */
+    private $transport;
+
+    public function __construct(TransportInterface $transport)
+    {
+        $this->transport = $transport;
+    }
+
+    public static function addVersionProperties(Node $node, TransportInterface $transport)
```
dbu: we should not have static methods

src/Jackalope/Version/VersionHandler.php
danrot: Should I add these constants there?

src/Jackalope/Version/VersionHandler.php
dbu: afaik we can manually set a uuid on a new node if we need to reference it before persisting it. that is what phpcr-odm does. maybe the same would help here too.

i am a bit confused by the code here, maybe splitting this into private methods would help: initVersioning for the general setup, and initVersionHistory for a specific node that has no history yet.

src/Jackalope/Version/VersionHandler.php
dbu: doing it on workspace initialization would make sense to me.

src/Jackalope/Version/VersionHandler.php
danrot: I have already tried to set the property like this:

```php
$versionHistory->setProperty('jcr:uuid', UUIDHelper::generateUUID());
```

But that didn't help...

src/Jackalope/Version/VersionHandler.php
dbu: i am quite uncertain. it would make most sense probably to translate the constant programmatically, with the current namespace registry. (there is a phpcr feature called local namespace remapping, so theoretically a user could remap `mix` to something else and all our code would break).

@dantleech do you have an idea for this? should we change Node::isNodeType so that it handles the long form of node type names too? sounds like the least surprising way of doing things.

dantleech: I guess making `isNodeType` namespace aware wouldn't be too difficult? @danrot would that help?

hacfi:
Awesome feature @danrot ! Would love to see this in Jackalope.

src/Jackalope/Version/VersionHandler.php
danrot: Yeah, that would definitely help :-)

src/Jackalope/Version/VersionHandler.php
danrot: @dbu Is it ok to make the `storeNode`-method public? I also need to create another node using the transport in the `VersionHandler`. Using the `storeNodes` method to save just one node.

danrot:
@hacfi I am doing my best :-) Actually working on it for my master thesis, so i should be finished at the end of june. But even I won't manage that for some reasons I will complete it later, it's really important to me ;-)

@dbu I am trying to implement some versioning logic in the `VersionHandler`, but it starts to feel a bit werid... I have already talked to @dantleech today about using the `ObjectManager` there, but he thought it might have a too big performance impact. On the other side I have the feeling that making the `storeNode` method of the client available in the interface would be a bad BC break, and if we really do that some validation might be skipped.

What's your opinion on that? Should I just use this stdClasses? Feels quite error prone to me... Or is it worth the performance overhead to use the ObjectManager? And if yes, what's the best way to get to the object manager in my handler class? Passing the factory somehow?

src/Jackalope/Version/VersionHandler.php
dbu: can't we just go through the object manager to create version history nodes? we also need to take care of session operations that only get persisted on session->save() vs immediate things...

danrot: I've just talked to @dantleech about this today, and we came to the conclusion that it would make more sense to not include this versioning stuff in the transport at all, since it should be the same for every storage layer (with jackrabbit being the exception, because the database is already handling versioning for us). So Dan thought it might be a good idea to replace/extend the `VersionManager` for Doctrine DBAL. Actually it would even make more sense to make the new one the default, which would also be valid for MongoDB and FS, and find a way to use a different one for Jackrabbit.

Dan also said that would split up the giant `Client` class, which he wants to split up in Jackalope 2.0

@dbu WDYT?

dbu: i would prefer to keep the architecture as it is now, but provide a handler that the client can delegate these things to, if the generic handler fits. that would keep things separated but allows to do backend specific performance tweaks.

the creation of infrastructure should also be in that handler rather than a VersionManager, or we will conflict with jackalope-jackrabbit which does not want to have any of this done on client side...

danrot: Ok, so keep the `VersionHandler`, but inject the `ObjectManager` to it? I tried to do that in the first place, but I didn't find a way to this easily, ergo without breaking the current architecture. The only option I've seen so far is to add something like the following after line https://github.com/jackalope/jackalope/blob/be3d34bd63cdd9200d9ec469dee0bc0cf671ccda/src/Jackalope/ObjectManager.php#L164

```
if ($this->transport implements MyVersioningInterface){
    $this->transport->setVersionHandler(new VersionHandler($this));
}
```

Of course I would have to think about a better name for this interface :smile:

dantleech: The problem with the existing architecture is that the Jackalope (PHPCR) VersionManager implementation delegates to the ObjectManager, which delegates to the (jackalope implementation) Transport, which then needs access to the ObjectManager again (for the versioning). 

I think the cleanest way is to implement the version manager using the PHPCR `VersionManager` interface -- and in that case it could either be use as a substitute for the Jackalope VersionManager implementation or it could be delegated to.

danrot: But the problem with how to inject the class implementing the `VersionManagerInterface` does still exist...

dbu: my idea is that the Client class would instantiate the helper to handle the methods of the version transport interface. much like we use the NodeProcessor.
checkin and checkout are direct operations, they do not need to go through the session. i don't think we need the ObjectManager here.

one thing we have to make sure is that the version subtree can not be manipulated through the normal node / property methods by the user of the library.

danrot: What else should we go trough? Should we use the methods from the transport?

Currently the transport is creating the `VersionHandler`, but I can only inject the transport interface there.

The signature of the checkin method just contains a path, and I couldn't find a solution to load a node by the path, and saving some information in a different node just using the transport, since everything I could find in the transport was using `Jackalope\Node` or `Jackalope\Property`. So even if I would like to don't use the `ObjectManager`, I would need it to inject it into the `Jackalope\Node` or `Jackalope\Property`, since their constructors requre an `ObjectManager` to be passed.

danrot:
Where would be the best place to add this `jcr:system/jcr:versionStorage`-node for the tests? I have just realized that the fixture for doctrine get generated on the fly, and adding these nodes to the fixtures in the phpcr-api-tests also feels dirty...

dantleech:
How does jackrabbit create the system nodes? Should these nodes be created automatically when the workspace is created / used for the first time?

danrot:
We already agreed that we would like to create these nodes when the workspace create command is executed, but as far as I have seen this command is not executed in the tests. For the tests there are simply some rows inserted in the database using the phpunit database extension.

PR #282, opened 2015-04-29
Enable collection of backtrace

dantleech:
This PR enables the collection of the `debug_backtrace`.

### DIFF
--- a/src/Jackalope/Transport/Logging/DebugStack.php
+++ b/src/Jackalope/Transport/Logging/DebugStack.php
@@ -28,6 +28,13 @@ class DebugStack implements LoggerInterface
     public $enabled = true;
 
     /**
+     * Show the debug backtrace for each call
+     *
+     * @var boolean
+     */
+    public $backtrace = false;
+
+    /**
      * @var float|null
      */
     public $start = null;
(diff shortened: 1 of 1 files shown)


src/Jackalope/Transport/Logging/DebugStack.php
```diff
@@ -28,6 +28,13 @@ class DebugStack implements LoggerInterface
     public $enabled = true;
 
     /**
+     * Show the debug backtrace for each call
+     *
+     * @var boolean
+     */
+    public $backtrace = false;
```
dbu: can this property be private?

dbu:
looks good otherwise. the test failure is with hhvm - i have seen this with other pull requests too, this is strange, looks like a hhvm regression to me.

src/Jackalope/Transport/Logging/DebugStack.php
dantleech: It could be. I was folllowing the example of `enabled` above, which I guess is `public` for convenience...

dbu: okay. enabled seems never used outside the class either. not sure, but i don't care too much here right now. merged so we can see if the bundle works too.

PR #283, opened 2015-05-02
Improve performance of StandardNodeTypes::getNodeTypeData

anthonysterling:
I added a static variable to hold the evaluated node type data. A
quick, simple, test shows an improvement from ~2.0153 to ~0.0011
for 10,000 calls.

```
$start = microtime(true);

for ($i = 0; $i < 10000; $i++) {
    StandardNodeTypes::getNodeTypeData();
}

printf("%s\n", microtime(true) - $start);
```

I don't use this library, so I may be missing something, but I
thought that this might be a quick win.

### DIFF
--- a/src/Jackalope/Transport/StandardNodeTypes.php
+++ b/src/Jackalope/Transport/StandardNodeTypes.php
@@ -26,3822 +26,3828 @@ class StandardNodeTypes
      */
     public static function getNodeTypeData()
     {
-        return array(
-            'mix:created' =>
-            array(
-                'name' => 'mix:created',
-                'isAbstract' => false,
-                'isMixin' => true,
-                'isQueryable' => true,
-                'hasOrderableChildNodes' => false,
-                'primaryItemName' => NULL,
-                'declaredSuperTypeNames' =>
-                array(
-                ),
-                'declaredPropertyDefinitions' =>
-                array(
-                    0 =>
-                    array(
-                        'declaringNodeType' => 'mix:created',
-                        'name' => 'jcr:createdBy',
-                        'isAutoCreated' => true,
-                        'isMandatory' => false,
-                        'isProtected' => true,
-                        'onParentVersion' => 1,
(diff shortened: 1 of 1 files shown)


lsmith77:
ok cool! I think I tried this as well today and it didn't seem to change anything but maybe I got confused as I was testing this inside an application and there might have been cache layers in between.

will check it out !

anthonysterling:
Brilliant, thanks Lukas.

lsmith77:
ok .. analyzed things further. we actually only call this method once in most cases I just realized, so this approach isn't going to help. so the question is how to most efficiently, load this array once in a request. note that I am using an opcode cache here.

now I also tried using Doctrine Cache with this and it turns out that it helps a bit (cutting the time in half) but is still surprisingly slow (still takes 300ms) with APCU being slightly faster than using the file system cache (which is based on php serialize).

in general this array can likely be simplified anyway as its essentially just a copy of the data structure we get from jackrabbit and we were just lazy and kept the same data structure for Doctrine DBAL. so maybe simplifying the data structure is maybe the better approach.

lsmith77:
Another approach would be to refactor the Jackalope NodeTypeManager to no longer simply read all node types at once but to actually only read the node types that are needed. This is probably a much more memory friendly approach as well.

dbu:
or could we apc-cache the actual objects instead of arrays? that would
benefit all jackalope flavors and probably would be the most efficient.

dbu:
is #287 an alternative to this? should we keep this still open?

lsmith77:
none of the approaches really convinced me if they really improve performance in the case of where we just do a single query in a request. I fear the only proper approach would be no longer loading all node types at once.

dbu:
so should we close the pull requests and keep an issue about the
refactoring for when/if a version 2 happens?

anthonysterling:
Going to close this as it's stale and probably out of date now. :)

PR #284, opened 2015-05-02
[POC] Lazy load property

dantleech:
This PR _experiments_ with lazy-loading properties -- do not review :)

### DIFF
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -745,7 +745,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         $this->checkState();
 
 
-        $names = self::filterNames($nameFilter, array_merge(array_keys($this->propertyData), array_keys($this->properties)));
+        $names = self::filterNames($nameFilter, $this->getPropertyNames());
 
         $result = array();
         foreach ($names as $name) {
@@ -766,7 +766,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         $this->checkState();
 
         // OPTIMIZE: do not create properties in constructor, go over array here
-        $names = self::filterNames($nameFilter, array_keys($this->propertyData));
+        $names = self::filterNames($nameFilter, $this->getPropertyNames());
         $result = array();
         foreach ($names as $name) {
             $property = $this->getProperty($name);
@@ -822,7 +822,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
     {
         $this->checkState();
 
-        if (isset($this->properties['jcr:uuid'])) {
+        if (isset($this->properties['jcr:uuid']) || isset($this->propertyData['jcr:uuid'])) {
             return $this->getPropertyValue('jcr:uuid');
         }
 
@@ -1680,5 +1680,11 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         }
 
         return $this->_setProperty($name, $value, $type, true);
+    }
+
+    private function getPropertyNames()
+    {
+        $names = array_unique(array_merge(array_keys($this->propertyData), array_keys($this->properties)));
+        return $names;
     }
 }
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -655,18 +655,22 @@ class Node extends Item implements IteratorAggregate, NodeInterface
     {
         $this->checkState();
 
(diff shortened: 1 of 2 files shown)


src/Jackalope/Node.php
```diff
@@ -886,7 +890,7 @@ public function hasProperty($relPath)
         $this->checkState();
 
         if (false === strpos($relPath, '/')) {
-            return isset($this->properties[$relPath]);
+            return isset($this->properties[$relPath]) || isset($this->propertyData[$relPath]);
```
dbu: do you even need to check $this->properties ?

dbu:
interesting! i could not resist and had a look at it. i think what you currently miss is that when new properties are added, you also need to add their names to the $propertiesData, or you will "forget" them in some cases.

src/Jackalope/Node.php
dantleech: no. .this was from an earlier stage .. :)

dantleech: Actually `->setProperty` sets `properties` directly.

src/Jackalope/Node.php
```diff
@@ -737,13 +744,13 @@ public function getProperties($nameFilter = null)
     {
         $this->checkState();
 
-        //OPTIMIZE: lazy iterator?
```
dbu: what you do below does not create a lazy iterator. but probably we don't need one.

src/Jackalope/Node.php
```diff
@@ -759,21 +766,22 @@ public function getPropertiesValues($nameFilter = null, $dereference = true)
         $this->checkState();
 
         // OPTIMIZE: do not create properties in constructor, go over array here
```
dbu: i think you just did that, so you can remove this comment.

dbu:
dont you need to add information to the raw data as well when creating a new property: https://github.com/jackalope/jackalope/blob/lazy_load_property/src/Jackalope/Node.php#L1546-L1561

and what about deleting a property that has not been loaded yet?
https://github.com/jackalope/jackalope/blob/lazy_load_property/src/Jackalope/Node.php#L571-L574 i think you need to force property instantiation here.

dantleech:
>> dont you need to add information to the raw data as well when creating a new property:

I didn't think so, as we always try to load first from `$properties` and only then fallback to the "raw data".

>> and what about deleting a property that has not been loaded yet?

maybe, the tests are still failing on the implementation(s) and I am not sure why.

dantleech:
Note to future self: `hasProperty` and `isset($this->properties) || isset($this->propertyData)` is not strictly equivilent, should add a separate `hasLocaleProperty` method to the node.

dantleech:
I have reduced the errors to 7 with 5 failures, it requires probably only a few fixes before it "works". But its a bit of a mess, and I am sure I will introduce regressions somehow with the current work. But maybe it would be interesting for benchmarking and seeing if its worth finishing off.

src/Jackalope/Node.php
```diff
@@ -942,12 +959,12 @@ public function getMixinNodeTypes()
...
         $this->checkState();
 
-        if (!isset($this->properties['jcr:mixinTypes'])) {
+        if (!$this->hasProperty('jcr:mixinTypes')) {
             return array();
         }
         $res = array();
         $ntm = $this->session->getWorkspace()->getNodeTypeManager();
-        foreach ($this->properties['jcr:mixinTypes']->getValue() as $type) {
+        foreach ($this->getPropertyValue('jcr:mixinTypes') as $type) {
```
dantleech: One tricky part is knowing when we can call these "API" methods, as they will always check the state and break thiings if the operation is "internal" and doesn't care about the external state.

src/Jackalope/Node.php
```diff
@@ -736,17 +743,21 @@ public function getPropertyValueWithDefault($relPath, $defaultValue)
     public function getProperties($nameFilter = null)
     {
         $this->checkState();
+        $names = self::filterNames($nameFilter, $this->getPropertyNames());
+        $result = $this->getLocalProperties($names);
+
+        return new ArrayIterator($result);
+    }
+
+    private function getLocalProperties($names = null)
```
dantleech: This is basically the same as `getProperties` but does not check the state.

dantleech:
Hmm, and a quick benchmark doesn't show any improvement currently, so I guess something is going wrong.

PR #285, opened 2015-05-05
Added VersionHandler with basic checkin and checkout methods

danrot:
This PR adds a VersionHandler, which can easily be reused among different transport layers. Currently there are very basic (and not yet complete) checkin and checkout methods.

PR #286, opened 2015-05-19
Adding version history support

danrot:
These changes will make the tests in the `VersionHistoryTest` working. It adds correct chaining of all the versions. I am skipping the tests for deleting versions, since deleting versions is also optional according to the specification.

This PR is based on the work in #285

https://github.com/jackalope/jackalope-doctrine-dbal/pull/282 is the corresponding PR for the doctrine dbal transport layer.

### DIFF
--- a/src/Jackalope/ObjectManager.php
+++ b/src/Jackalope/ObjectManager.php
@@ -1809,19 +1809,18 @@ class ObjectManager
 
     /**
      * Updates a single node via the transport layer if the node is modified
+     *
      * @param NodeInterface $node The node to update
      */
-    private function updateNode($node)
+    private function updateNode(NodeInterface $node)
     {
-        /** @var $node Node */
-        if ($node->isModified()) {
-            if (!$node instanceof NodeInterface) {
-                throw new RepositoryException('Internal Error: Unknown type ' . get_class($node));
-            }
-            $this->transport->updateProperties($node);
-            if ($node->needsChildReordering()) {
-                $this->transport->reorderChildren($node);
-            }
+        if (!$node->isModified()) {
+            return;
+        }
+
+        $this->transport->updateProperties($node);
+        if ($node->needsChildReordering()) {
+            $this->transport->reorderChildren($node);
         }
     }
 }
--- a/src/Jackalope/ObjectManager.php
+++ b/src/Jackalope/ObjectManager.php
@@ -5,6 +5,7 @@ use ArrayIterator;
 use InvalidArgumentException;
 
 use Jackalope\Transport\NodeTypeFilterInterface;
+use Jackalope\Version\GenericVersioningInterface;
 use PHPCR\SessionInterface;
 use PHPCR\NodeInterface;
 use PHPCR\PropertyInterface;
@@ -820,16 +821,15 @@ class ObjectManager
             // loop through cached nodes and commit all dirty and set them to clean.
             if (isset($this->objectsByPath['Node'])) {
                 foreach ($this->objectsByPath['Node'] as $node) {
-                    /** @var $node Node */
-                    if ($node->isModified()) {
-                        if (! $node instanceof NodeInterface) {
-                            throw new RepositoryException('Internal Error: Unknown type '.get_class($node));
-                        }
-                        $this->transport->updateProperties($node);
-                        if ($node->needsChildReordering()) {
-                            $this->transport->reorderChildren($node);
-                        }
-                    }
+                    $this->updateNode($node);
+                }
+            }
+
+            if (isset($this->objectsByPath['Version\\Version'])
+                && $this->transport instanceof GenericVersioningInterface
+            ) {
+                foreach ($this->objectsByPath['Version\\Version'] as $node) {
+                    $this->updateNode($node);
                 }
             }
 
@@ -1805,5 +1805,23 @@ class ObjectManager
         // if the node moved away from this node, we did not find it in
         // objectsByPath and the calling parent node can forget it
         return true;
+    }
+
+    /**
+     * Updates a single node via the transport layer if the node is modified
+     * @param NodeInterface $node The node to update
+     */
+    private function updateNode($node)
+    {
+        /** @var $node Node */
+        if ($node->isModified()) {
+            if (!$node instanceof NodeInterface) {
+                throw new RepositoryException('Internal Error: Unknown type ' . get_class($node));
+            }
+            $this->transport->updateProperties($node);
+            if ($node->needsChildReordering()) {
+                $this->transport->reorderChildren($node);
+            }
+        }
     }
 }
--- a/src/Jackalope/Version/VersionHandler.php
+++ b/src/Jackalope/Version/VersionHandler.php
@@ -115,7 +115,6 @@ class VersionHandler
             );
         }
 
-        /** @var NodeInterface $baseVersionNode */
         $baseVersionUuid = $node->getProperty('jcr:baseVersion')->getString();
         $baseVersionNode = $this->objectManager->getNodeByIdentifier($baseVersionUuid, 'Version\Version');
 
@@ -135,7 +134,6 @@ class VersionHandler
 
         // FIXME add some kind of sharding
         // should avoid to have too many nodes on the same level
-        /** @var NodeInterface $versionNode */
         $versionNode = $versionHistoryNode->addNode(UUIDHelper::generateUUID(), 'nt:version');
         $versionNode->setProperty('jcr:uuid', UUIDHelper::generateUUID());
         $versionNode->setProperty('jcr:created', new \DateTime());
@@ -223,10 +221,10 @@ class VersionHandler
     }
 
     /**
-     * Creates a frozen node
+     * Attaches a nt:frozenNode copy of $node to $versionNode
      *
-     * @param NodeInterface $versionNode
-     * @param NodeInterface $node
+     * @param NodeInterface $versionNode The version to attach the nt:frozenNode to
+     * @param NodeInterface $node The node to be copied into an nt:frozenNode
      */
     private function createFrozenNode(NodeInterface $versionNode, NodeInterface $node)
     {
--- a/src/Jackalope/Version/VersionHandler.php
+++ b/src/Jackalope/Version/VersionHandler.php
@@ -116,7 +116,8 @@ class VersionHandler
         }
 
         /** @var NodeInterface $baseVersionNode */
-        $baseVersionNode = $node->getPropertyValue('jcr:baseVersion');
+        $baseVersionUuid = $node->getProperty('jcr:baseVersion')->getString();
+        $baseVersionNode = $this->objectManager->getNodeByIdentifier($baseVersionUuid, 'Version\Version');
 
         if (!$node->isCheckedOut()) {
             return $baseVersionNode->getPath();
@@ -129,7 +130,8 @@ class VersionHandler
         // TODO set subgraph to read only
 
         /** @var NodeInterface $versionHistoryNode */
-        $versionHistoryNode = $node->getPropertyValue('jcr:versionHistory');
+        $versionHistoryUuid = $node->getProperty('jcr:versionHistory')->getString();
+        $versionHistoryNode = $this->objectManager->getNodeByIdentifier($versionHistoryUuid, 'Version\VersionHistory');
 
         // FIXME add some kind of sharding
         // should avoid to have too many nodes on the same level
--- a/src/Jackalope/Version/VersionHandler.php
+++ b/src/Jackalope/Version/VersionHandler.php
@@ -2,6 +2,7 @@
 
 namespace Jackalope\Version;
 
+use Jackalope\Property;
 use Jackalope\Session;
 use Jackalope\Transport\AddNodeOperation;
 use Jackalope\Transport\WritingInterface;
@@ -76,8 +77,9 @@ class VersionHandler
         $additionalOperations[] = new AddNodeOperation($versionLabels->getPath(), $versionLabels);
         $rootVersion = $versionHistory->addNode('jcr:rootVersion', 'nt:version');
         $rootVersion->setProperty('jcr:uuid', UUIDHelper::generateUUID());
-        $rootVersion->setProperty('jcr:predecessors', array());
-        $rootVersion->setProperty('jcr:successors', array()); // not part of the specification, but seems to be required
+        $rootVersion->setProperty('jcr:predecessors', array(), PropertyType::REFERENCE);
+        // not part of the specification, but seems to be required
+        $rootVersion->setProperty('jcr:successors', array(), PropertyType::REFERENCE);
         $additionalOperations[] = new AddNodeOperation($rootVersion->getPath(), $rootVersion);
 
         // TODO add frozen node to root version
@@ -148,26 +150,24 @@ class VersionHandler
             $node->getProperty('jcr:predecessors')->getString(),
             PropertyType::REFERENCE
         );
+        $versionNode->setProperty(
+            'jcr:successors',
+            array(),
+            PropertyType::REFERENCE,
+            false
+        );
 
         $node->setProperty(
             'jcr:predecessors',
             array($versionNode),
             PropertyType::REFERENCE,
             false
-        ); // TODO double check with specification, this line seems not to be part of it
+        );
+
         foreach ($versionNode->getPropertyValueWithDefault('jcr:predecessors', array()) as $predecessorNode) {
-            $predecessorNode->setProperty(
-                'jcr:successors',
-                // TODO check why array_unique is required
-                array_unique(
-                    array_merge(
-                        $predecessorNode->getProperty('jcr:successors')->getString(),
-                        array($versionNode->getIdentifier())
-                    )
-                ),
-                PropertyType::REFERENCE,
-                false
-            );
+            /** @var NodeInterface $predecessorNode */
+            $successorProperty = $predecessorNode->getProperty('jcr:successors');
+            $successorProperty->addValue($versionNode);
             $predecessorNode->setModified();
         }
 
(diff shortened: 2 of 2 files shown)


src/Jackalope/Version/VersionHandler.php
```diff
@@ -0,0 +1,224 @@
...
+        $frozenNode->setProperty('jcr:frozenPrimaryType', $node->getProperty('jcr:primaryType'));
+        if ($frozenNode->hasProperty('jcr:mixinTypes')) {
+            $frozenNode->setProperty('jcr:frozenMixinTypes', $node->getProperty('jcr:mixinTypes'));
+        }
+        $frozenNode->setProperty('jcr:frozenUuid', $node->getProperty('jcr:uuid'));
+
+        /** @var NodeInterface $baseVersionNode */
+        $baseVersionNode = $node->getPropertyValue('jcr:baseVersion');
+        $baseVersionNode->setProperty('jcr:successors', array($versionNode), PropertyType::REFERENCE, false);
+        $baseVersionNode->setModified();
```
danrot: I find myself doing this quite often, since `setProperty` doesn't handle the state at all, if it is called with `$validate = false`. I am not sure if this is ok this way, or we should change the behavior of the `setProperty`-method resp. the subsequent function calls. /cc @dbu @lsmith77 @dantleech

src/Jackalope/Version/VersionHandler.php
```diff
@@ -0,0 +1,225 @@
...
+        // TODO create frozen node
+        $frozenNode = $versionNode->addNode('jcr:frozenNode', 'nt:frozenNode');
+        $frozenNode->setProperty('jcr:frozenPrimaryType', $node->getProperty('jcr:primaryType'));
+        if ($frozenNode->hasProperty('jcr:mixinTypes')) {
+            $frozenNode->setProperty('jcr:frozenMixinTypes', $node->getProperty('jcr:mixinTypes'));
+        }
+        $frozenNode->setProperty('jcr:frozenUuid', $node->getProperty('jcr:uuid'));
+
+        $baseVersionNode->setProperty('jcr:successors', array($versionNode), PropertyType::REFERENCE, false);
+        $baseVersionNode->setModified();
```
danrot: Same comment as https://github.com/jackalope/jackalope/pull/286#discussion_r30601250, it just got lost due to an update. Do not know if it is a good idea to set this on my own.

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -161,12 +174,15 @@ private function processNodeWithType(NodeInterface $node, NodeType $nodeTypeDefi
             }
 
             if (!$node->hasProperty($propertyDef->getName())) {
-                if ($propertyDef->isMandatory() && !$propertyDef->isAutoCreated()) {
+                if ($propertyDef->isMandatory()
+                    && !$propertyDef->isAutoCreated()
+                    && $nodeTypeDefinition->getName() !== VersionHandler::MIX_VERSIONABLE
```
dantleech: Maybe move this to a `private isRequired()` method

````php
if ($this->propertyIsRequired($property)) {
     throw ...
}
````

Comment (danrot) [inline src/Jackalope/NodeType/NodeProcessor.php]: That code is actually already in #285... I don't think that it will be too easy to review this one before the other one is merged :-) But I can change this if the others agree.

**Comment by danrot** [comment]
@dbu @dantleech Wasn't there already a check in jackalope/jackalope for the `jcr:isCheckedOut`-property? I was quite sure I've had the `testWriteNotCheckedOutVersion` test already running, but it's failing at the moment, and I can't find this check in the code at all...

File: src/Jackalope/ObjectManager.php [hunk 4]
```diff
@@ -818,12 +818,12 @@ public function save()
             $this->executeOperations($this->operationsLog);
 
             // loop through cached nodes and commit all dirty and set them to clean.
-            if (isset($this->objectsByPath['Node'])) {
-                foreach ($this->objectsByPath['Node'] as $node) {
+            foreach ($this->objectsByPath as $nodes) {
```
dbu: are you aware that the first level of index is the type of nodes? (e.g. with jackrabbit, there is also a VersionNode subpart in objectsByPath)

other type of objects than 'Node' should not be handled in this loop. i guess this whole object store is something we really want to clean up in 2.0 /cc @dantleech

src/Jackalope/Version/VersionHandler.php
```diff
@@ -74,12 +76,14 @@ public function addVersionProperties(NodeInterface $node)
         $additionalOperations[] = new AddNodeOperation($versionLabels->getPath(), $versionLabels);
         $rootVersion = $versionHistory->addNode('jcr:rootVersion', 'nt:version');
         $rootVersion->setProperty('jcr:uuid', UUIDHelper::generateUUID());
+        $rootVersion->setProperty('jcr:predecessors', array());
+        $rootVersion->setProperty('jcr:successors', array()); // not part of the specification, but seems to be required
```
dbu: we need to check this with jackalope-jackrabbit to make sure it has no undesired side effects.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -103,14 +107,20 @@ public function checkinItem($path)
...
+                sprintf(
+                    'Node "%s" contains unsaved changes',
+                    $path
+                )
+            );
         }
 
+        /** @var NodeInterface $baseVersionNode */
+        $baseVersionUuid = $node->getProperty('jcr:baseVersion')->getString();
+        $baseVersionNode = $this->objectManager->getNodeByIdentifier($baseVersionUuid, 'Version\Version');
```
dbu: @dantleech for 2.0 it would be really nice if the transport could discover that this is a version instead of we having to tell it explicitly. then these two lines could read `$baseVersionNode = $node->getPropertyValue('jcr:baseVersion');`

src/Jackalope/Version/VersionHandler.php
```diff
@@ -103,14 +107,20 @@ public function checkinItem($path)
...
-            ));
+            throw new InvalidItemStateException(
+                sprintf(
+                    'Node "%s" contains unsaved changes',
+                    $path
+                )
+            );
         }
 
+        /** @var NodeInterface $baseVersionNode */
```
dbu: afaik this is already known from the return type annotation on ObjectManager::getNodeByIdentifier.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -119,27 +129,63 @@ public function checkinItem($path)
...
         // TODO set subgraph to read only
 
-        $versionHistoryNode = $node->getPropertyValue('jcr:versionHistory');
+        /** @var NodeInterface $versionHistoryNode */
+        $versionHistoryUuid = $node->getProperty('jcr:versionHistory')->getString();
+        $versionHistoryNode = $this->objectManager->getNodeByIdentifier($versionHistoryUuid, 'Version\VersionHistory');
 
         // FIXME add some kind of sharding
         // should avoid to have too many nodes on the same level
+        /** @var NodeInterface $versionNode */
```
dbu: i think this is unneeded, NodeInterface::addNode already says this in the return type

src/Jackalope/Version/VersionHandler.php
```diff
@@ -174,4 +221,40 @@ public function checkoutItem($path)
 
         $this->session->save();
     }
+
+    /**
+     * Creates a frozen node
```
dbu: can you be more explicit here? "Attaches a nt:frozenNode copy of $node to $versionNode"

src/Jackalope/Version/VersionHandler.php
```diff
@@ -174,4 +221,40 @@ public function checkoutItem($path)
...
+     *
+     * @param NodeInterface $versionNode
+     * @param NodeInterface $node
```
dbu: The node to be copied into an nt:frozenNode

src/Jackalope/Version/VersionHandler.php
```diff
@@ -174,4 +221,40 @@ public function checkoutItem($path)
 
         $this->session->save();
     }
+
+    /**
+     * Creates a frozen node
+     *
+     * @param NodeInterface $versionNode
```
dbu: The version to attach the frozen node to.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -119,27 +129,63 @@ public function checkinItem($path)
...
+            PropertyType::REFERENCE
+        );
 
-        $versionNode->setProperty('jcr:predecessors', $node->getProperty('jcr:predecessors')->getString());
-        $node->setProperty('jcr:predecessors', array(), PropertyType::REFERENCE, false);
-        foreach ($versionNode->getProperty('jcr:predecessors') as $predecessorUuid) {
-            $predecessor = $this->objectManager->getNodeByIdentifier($predecessorUuid);
-            $predecessor->setProperty('jcr:successors', array_merge($predecessor->getPropertyValueWithDefault('jcr:successors', array()), array($versionNode)));
+        // FIXME Find a better way to solve the referencing not existing nodes issue
+        $this->session->save();
```
dbu: can't you somehow create the uuid already so you can reference them?

dbu:
i did not find anything on an isCheckedOut check. but lets keep write protection separate from this PR.

this code is shared with jackalope-jackrabbit. we need to find a way to only execute the creation of version things when we have the client side version handling. is that already the case?

and i think we indeed need a better solution for the object store. i think for 2.0 we should look into what ocramius did. but for now, is there a chance you can be explicit what class you expect to operate on?

src/Jackalope/ObjectManager.php
danrot: Ah, ok... but I need to save the version nodes... Maybe there should be a second option for workspace writes? (since somebody has to set the correct values on the version nodes...)

src/Jackalope/Version/VersionHandler.php
danrot: The `VersionHandler` won't be instantiated for jackalope-jackrabbit, so this shouldn't be an issue.

danrot:
@dbu Yeah, the `VersionHandler` is only instantiated if the Client is implementing the `GenericVersioningInterface`, so this should not be an issue.

src/Jackalope/Version/VersionHandler.php
dbu: good then!

src/Jackalope/ObjectManager.php
dbu: lets add a loop for saving all version things if the transport is a GenericVersionInterface or what its called. that should loop over Version/Version and whatever other keys are created by the VersionHandler adding non-Node classes.

dbu: maybe just something like `if ($this->transport instanceof ...) {$this->saveVersionNodes();}` where you implement that loop. or move the loop into a method with the class as parameter, to avoid duplication. then always call it with Node and possibly with Version and so on.

danrot: Would jackrabbit allow this? I thought the version nodes should not be updated manually...

src/Jackalope/Version/VersionHandler.php
danrot: Tried to remove it, but it didn't work, although I have set a UUID on the node to be referenced...

danrot:
@dbu Removed the write protection from the ImplementationLoader in https://github.com/jackalope/jackalope-doctrine-dbal/pull/282, so that I can do that in a separate PR.

> and i think we indeed need a better solution for the object store. i think for 2.0 we should look into what ocramius did. but for now, is there a chance you can be explicit what class you expect to operate on?

I am not sure what you mean by that, but is it what I have done in https://github.com/jackalope/jackalope/commit/10c54f8d5df2b63f6ffd2c7276c46eec08051c31?

If yes this should also be fine to merge :grinning:

src/Jackalope/Version/VersionHandler.php
danrot: Found another bug, now I was also able to remove this save process...

src/Jackalope/ObjectManager.php
```diff
@@ -1806,4 +1806,22 @@ public function purgeDisappearedNode($absPath, $keepChanges)
...
         return true;
     }
+
+    /**
+     * Updates a single node via the transport layer if the node is modified
+     * @param NodeInterface $node The node to update
+     */
+    private function updateNode($node)
+    {
+        /** @var $node Node */
```
dbu: lets put the typehint into the method. then we can drop this comment, and also the weird instanceof check below. its only a sanity check - the typehint is better for that.

src/Jackalope/Version/VersionHandler.php
dbu: i think you need to explicitly specify the type of this property, or it will be considered a type string property.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -119,27 +128,52 @@ public function checkinItem($path)
...
+        $baseVersionNode->setModified();
+        $versionNode->setProperty(
+            'jcr:predecessors',
+            $node->getProperty('jcr:predecessors')->getString(),
+            PropertyType::REFERENCE
+        );
+
+        $node->setProperty(
+            'jcr:predecessors',
+            array($versionNode),
```
dbu: is this only ever the nearest predecessors, not all?

src/Jackalope/Version/VersionHandler.php
```diff
@@ -119,27 +128,52 @@ public function checkinItem($path)
...
+            PropertyType::REFERENCE,
+            false
+        ); // TODO double check with specification, this line seems not to be part of it
+        foreach ($versionNode->getPropertyValueWithDefault('jcr:predecessors', array()) as $predecessorNode) {
+            $predecessorNode->setProperty(
```
dbu: can't you just append to the property with Property::addValue?

src/Jackalope/ObjectManager.php
danrot: I was not sure if you rely on this exception being thrown somewhere, that's why I just extracted it. Will change it that way.

dbu: i see. but i hope not that we relied on that. its an internal error, it
should not happen.

src/Jackalope/Version/VersionHandler.php
danrot: What do you mean by all? I think you can only have more than one predecessor if you have branching and merging, which is currently not implemented.

src/Jackalope/Version/VersionHandler.php
danrot: Didn't know about this function, will try.

src/Jackalope/Version/VersionHandler.php
dbu: ah okay. yeah, i think you are right, its this linked list concept.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -174,4 +209,40 @@ public function checkoutItem($path)
...
+            ) {
+                continue;
+            }
+
+            $onParentValue = $property->getDefinition()->getOnParentVersion();
+            if ($onParentValue != OnParentVersionAction::COPY && $onParentValue != OnParentVersionAction::VERSION) {
+                continue;
+            }
+
+            // TODO apply other steps based on onParentValue
```
dbu: is this a TODO for this pull request or for later?

src/Jackalope/ObjectManager.php
```diff
@@ -1806,4 +1806,22 @@ public function purgeDisappearedNode($absPath, $keepChanges)
...
+        if ($node->isModified()) {
```
dantleech: You could return early here.

````php
if (false === $node->isModified()) {
    return;
}
````

Comment (dbu) [inline src/Jackalope/ObjectManager.php]: or just `!$node->isModified()`

File: src/Jackalope/Version/VersionHandler.php (hunk 16 above)
Comment (danrot) [inline src/Jackalope/Version/VersionHandler.php]: It's for later. Actually it looks to me like there should be done more according to the specification, but the phpcr-api-tests do not cover these cases. I thought I'll leave the todos here, so that later the phpcr-api-tests can be adapted accordingly.

Comment (dbu) [inline src/Jackalope/Version/VersionHandler.php]: okay. can you please open an issue on phpcr-api-tests to remind us?

Comment (danrot) [inline src/Jackalope/Version/VersionHandler.php]: https://github.com/phpcr/phpcr-api-tests/issues/159

File: src/Jackalope/Version/VersionHandler.php [hunk 18]
```diff
@@ -119,27 +130,50 @@ public function checkinItem($path)
...
+            PropertyType::REFERENCE,
+            false
+        );
+
+        $node->setProperty(
+            'jcr:predecessors',
+            array($versionNode),
+            PropertyType::REFERENCE,
+            false
+        ); // TODO double check with specification, this line seems not to be part of it
```
dbu: did you double check? does it make sense? would be nice if we can resolve this TODO - otherwise i fear it won't be adressed.

dbu:
apart from that TODO i commented, this seems good to me now

src/Jackalope/Version/VersionHandler.php
danrot: It's not part of the specification, but definitely makes sense. The spec says that you should copy the predecessors from the actual node, which will never have a predecessor without this statement. So i guess I just remove the comment, if this is ok for you.

dbu: :+1:

danrot:
@dbu That's fixed.

dbu:
:tada:

danrot:
@dbu previously on this thread you have mentioned that we should have a look at what @Ocramius did for the ObjectStore in 2.0. Can you elaborate in more detail what you have been talking about here? Might be interesting for my thesis.

dbu:
tried to find it but found no git repo - @ocramius or @dantleech might point you to where this is. but at this point its just ideas. but could be something for the future ideas.

PR #287, opened 2015-05-27
[WIP] improve performance for loading the standard node types

lsmith77:
also requires changes in the transports

essentially the issue is that loading the standard node types is super slow, yet we load these everytime we run a query with the doctrine dbal transport

### DIFF
--- a/src/Jackalope/NodeType/NodeTypeManager.php
+++ b/src/Jackalope/NodeType/NodeTypeManager.php
@@ -100,22 +100,29 @@ class NodeTypeManager implements IteratorAggregate, NodeTypeManagerInterface
      *
      * On fetch all, already cached node types are kept.
      *
-     * @param string $name type name to fetch. defaults to null which will
+     * @param array $name type name to fetch. defaults to null which will
      *      fetch all nodes.
      */
-    protected function fetchNodeTypes($name = null)
+    protected function fetchNodeTypes($names = null)
     {
         if ($this->fetchedAllFromBackend) {
             return;
         }
 
-        if (null !== $name) {
-            if (!empty($this->primaryTypes[$name]) || !empty($this->mixinTypes[$name])) {
-                return; //we already know this node
+        if (null !== $names) {
+            $missing = array();
+            foreach ($names as $name) {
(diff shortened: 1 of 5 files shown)


src/Jackalope/NodeType/NodeTypeManager.php
```diff
@@ -100,22 +100,29 @@ public function __construct(
      *
      * On fetch all, already cached node types are kept.
      *
-     * @param string $name type name to fetch. defaults to null which will
+     * @param array $name type name to fetch. defaults to null which will
```
dbu: this should be `$names` now

src/Jackalope/ObjectManager.php
```diff
@@ -654,6 +654,26 @@ public function getBinaryStream($path)
         return $this->transport->getBinaryStream($this->getFetchPath($path, 'Node'));
     }
 
+    public function hasNodeType($nodeTypeName)
+    {
+        return $this->transport->hasNodeType($nodeTypeName);
+
+        try {
```
dbu: this will never be executed, right?

src/Jackalope/Transport/TransportInterface.php
```diff
@@ -310,6 +310,10 @@ public function getWeakReferences($path, $name = null);
      */
     public function setNodeTypeManager($nodeTypeManager);
 
+    public function hasNodeType($nodeTypeName);
```
dbu: can we add phpdoc for this?

dbu: also, who will be responsible for caching this? afaik the transport never caches anything, so the OM should probably cache the check result. also, the NodeTypeManager would know if it already loaded the type so it could immediately return true - maybe the method should stay on NodeTypeManager and we give the NTM access to the transport?

src/Jackalope/ObjectManager.php
lsmith77: yeah .. I mainly still left it in .. since we actually have to think a bit more about the transport API since I have not prototyped what changes are needed in the Jackrabbit transport for this.

dbu:
there is #342 which improves for user node types. pavel claims that the speed for standard node types is very high on php 7 - should we just drop these changes?

Issue #210, opened 2014-03-19
null values in multivalued properties


uwej711:
If you try to store a null value in a mutlivalued property jackalope throws an exception when it tries to convert the null value to the type of the property. According to the spec null values in multivalued fields should be treated the same as null values for properties, i.e. remove the value from the multivalued property (as properties with value null are removed as well). See http://www.day.com/specs/jcr/2.0/10_Writing.html, "10.4.2.5 Multi-value Properties and Null".

* [ ] Create phpcr-api-tests test that stores multivalue with null
* [ ] Fix the issue - i guess right in the Property class on setValue

dbu:
started working on this, but its more tricky than i thought.

dbu:
started working on this in #288

dbu:
i merged the fix for this one. @uwej711 glad if you can check if it works as expected and i did not break anything in the process.

PR #288, opened 2015-06-09
cleanup property creation and handle null values in multivalue

dbu:
fix https://github.com/jackalope/jackalope/issues/210

as demonstrated in https://github.com/phpcr/phpcr-api-tests/pull/158

this was a huge mess, and the code is still really overcomplicated. separation between Node and Property seems not fully good either. but the tests are all green with jackalope-jackrabbit now.

input welcome - maybe we can do better in 2.0 / cc @dantleech

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "jackalope/jackalope-transport": "*",
         "php": ">=5.3.3",
         "ext-xml": "*",
-        "phpcr/phpcr-utils": "~1.1",
+        "phpcr/phpcr-utils": "~1.2,>=1.2.6",
         "phpcr/phpcr": "~2.1.0,>=2.1.0-beta12"
     },
     "provide": {
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -6,6 +6,7 @@ use ArrayIterator;
 use IteratorAggregate;
 use Exception;
 use InvalidArgumentException;
+use Jackalope\NodeType\NodeType;
 use LogicException;
 
 use PHPCR\PropertyType;
@@ -256,7 +257,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
                                     ? $rawData->{':' . $key}
                                     : PropertyType::valueFromName($rawData->{':' . $key});
                         } else {
-                            $type = $this->valueConverter->determineType(is_array($value) ? reset($value) : $value);
+                            $type = $this->valueConverter->determineType($value);
                         }
                         $this->_setProperty($key, $value, $type, true);
                         break;
@@ -529,12 +530,19 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         }
 
         if ($validate) {
+            if (is_array($value)) {
+                foreach ($value as $key => $v) {
+                    if (null === $v) {
+                        unset($value[$key]);
+                    }
+                }
+            }
             $types = $this->getMixinNodeTypes();
             array_push($types, $this->getPrimaryNodeType());
             if (null !== $value) {
                 $exception = null;
                 foreach ($types as $nt) {
-                    /** @var $nt \Jackalope\NodeType\NodeType */
+                    /** @var $nt NodeType */
                     try {
                         $nt->canSetProperty($name, $value, true);
                         $exception = null;
--- a/src/Jackalope/NodeType/NodeType.php
+++ b/src/Jackalope/NodeType/NodeType.php
@@ -182,7 +182,7 @@ class NodeType extends NodeTypeDefinition implements NodeTypeInterface
     {
         $propDefs = $this->getPropertyDefinitions();
         try {
-            $type = $this->valueConverter->determineType(is_array($value) ? reset($value) : $value);
+            $type = $this->valueConverter->determineType($value);
         } catch (ValueFormatException $e) {
             if ($throw) {
                 throw new ValueFormatException(sprintf('Invalid value for property "%s": %s', $propertyName, $e->getMessage()), $e->getCode(), $e);
--- a/src/Jackalope/Property.php
+++ b/src/Jackalope/Property.php
@@ -64,11 +64,12 @@ class Property extends Item implements IteratorAggregate, PropertyInterface
      * @var boolean
      */
     protected $isMultiple = false;
+
     /**
      * the type constant from PropertyType
      * @var int
      */
-    protected $type;
+    protected $type = PropertyType::UNDEFINED;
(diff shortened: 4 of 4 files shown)


src/Jackalope/NodeType/NodeType.php
```diff
@@ -182,7 +182,11 @@ public function canSetProperty($propertyName, $value, $throw = false)
     {
         $propDefs = $this->getPropertyDefinitions();
         try {
-            $type = $this->valueConverter->determineType(is_array($value) ? reset($value) : $value);
+            if (is_array($value) && 0 === count($value)) {
+                $type = PropertyType::STRING;
```
dantleech: So does this mean that empty arrays become empty strings?

src/Jackalope/Property.php
```diff
@@ -709,25 +685,44 @@ public function _setValue($value, $type = PropertyType::UNDEFINED)
          */
 
         if (PropertyType::UNDEFINED === $type) {
-            $type = $this->valueConverter->determineType(is_array($value) ? reset($value) : $value);
+            if (is_array($value) && 0 === count($value)) {
+                $type = PropertyType::STRING;
```
dantleech: This repeats the logic from before .. can it / would it make sense for it to be moved into the value converter?

src/Jackalope/Property.php
```diff
@@ -709,25 +685,44 @@ public function _setValue($value, $type = PropertyType::UNDEFINED)
...
         }
+        */
 
-        $value = $this->valueConverter->convertType($value, $targettype, $type);
+        if (PropertyType::BINARY !== $targettype || $constructor && $this->isNew()) {
+            // When in constructor mode, force conversion to re-determine the type as the desired type might not match the value
+            $value = $this->valueConverter->convertType($value, $targettype, $constructor ? PropertyType::UNDEFINED : $type);
+        }
 
         if (PropertyType::BINARY === $targettype) {
```
dantleech: Might as well `s/targettype/targetType`

src/Jackalope/Property.php
```diff
@@ -709,25 +685,44 @@ public function _setValue($value, $type = PropertyType::UNDEFINED)
...
+                $type = PropertyType::STRING;
+            } else {
+                $type = $this->valueConverter->determineType(is_array($value) ? reset($value) : $value);
+            }
         }
 
+        $targettype = $type;
+
+        /*
+         * TODO: find out with node type definition if the new type is allowed
```
dantleech: So currently this is not validated at all? or is it just not validated early?

src/Jackalope/NodeType/NodeType.php
dbu: i am not sure about this one. but when the array is empty and the type not specified, we need to chose something arbitrarily as there is no value to decide upon...

src/Jackalope/Property.php
dbu: i fear its only validated by the jackalope-jackrabbit backend.

this is not because of this pull request however, i only cleaned up that segment which was commented partially to do a noop.

src/Jackalope/Property.php
dbu: the problem is that when its an array, we `reset` which is an error on an empty array. we could make the value converter accept arrays as well and let it do the logic. will do that.

dbu: ah, i know why i did not: this is in the phpcr-utils, further complicating everything.

i just wanted to fix an edge case, not refactor everything :-( will do a PR on phpcr-utils and bump its version

composer.json
```diff
@@ -19,7 +19,7 @@
         "jackalope/jackalope-transport": "*",
         "php": ">=5.3.3",
         "ext-xml": "*",
-        "phpcr/phpcr-utils": "~1.1",
+        "phpcr/phpcr-utils": "~1.3",
```
dbu: if we consider https://github.com/phpcr/phpcr-utils/pull/153 a bugfix, this could be `~1.2, >=1.2.6`

dbu:
phpcr-utils side is now merged so this could theoretically be ready.

@lsmith77 can you have a good look at this please? might well be i messed up something.

for 2.0 we really should aim to be more explicit with binary value versus length, that is the culprit for half of that weird complexity (and hours of me trying to fix things until tests where green again)

src/Jackalope/Property.php
```diff
@@ -106,7 +107,7 @@ public function __construct(FactoryInterface $factory, array $data, $path, Sessi
 
         $this->wrapBinaryStreams = $session->getRepository()->getDescriptor(Repository::JACKALOPE_OPTION_STREAM_WRAPPER);
 
-        if (empty($data) && $new) {
+        if (is_null($data) && $new) {
```
lsmith77: a bit further down you changed the code from ``is_null`` to ``null ===`` ..

src/Jackalope/Node.php
```diff
@@ -529,12 +530,19 @@ public function setProperty($name, $value, $type = PropertyType::UNDEFINED, $val
         }
 
         if ($validate) {
+            if (is_array($value)) {
+                foreach ($value as $key => $v) {
```
lsmith77: you added more or less the same logic also to Propert::_setValue()

lsmith77:
I have added some comments from things I saw but for correctness, I sort of have to trust the tests.

src/Jackalope/Node.php
dbu: yes i know. but i need to do this before validating as well. i won't say its not a mess, but i feel its not worth doing much more refactoring if the end result will still be a mess.

the whole constructor and setup and validation logic needs to be remade for 2.0 i think.

dbu:
updated the PR. i will be in holidays now, can't promise to update other things soon. but i would keep that bit of code duplication (in the sum its a lot less duplication than before)

lsmith77:
looks good to me. I guess we should merge the test PR first, or adjust this PR to use the test branch?

dbu:
jackalope/jackalope does not run the api tests. and the transports lock to
a specific tag of the api tests. i ran the tests locally - i think in this
case this is the easiest to do.
Am 15.06.2015 11:16 schrieb "Lukas Kahwe Smith" <notifications@github.com>:

lsmith77:
this change caused Doctrine DBAL to fail https://travis-ci.org/jackalope/jackalope-doctrine-dbal/builds/72686710

dbu:
damn. i will have a look

dbu:
see #293

PR #289, opened 2015-06-11
Adding restore functionality for versioning support

danrot:
This PR adds restoring capabilities to the `VersionHandler`.

In order to use this functionality with jackalope-doctrine-dbal https://github.com/jackalope/jackalope-doctrine-dbal/pull/283 also needs to be merged.


src/Jackalope/Version/VersionHandler.php
```diff
@@ -246,4 +246,60 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+        }
+
+        $versionNode = $this->objectManager->getNodeByPath($versionPath, 'Version\Version');
+        $frozenNode = $versionNode->getNode('jcr:frozenNode');
+
+        // TODO reset primary type
+
+        if ($frozenNode->hasProperty('jcr:frozenMixinTypes')) {
+            $node->setProperty('jcr:mixinTypes', $frozenNode->getPropertyValue('jcr:frozenMixinTypes'));
+        }
```
dbu: is that property not always set? if not, you would need to delete the mixinTypes from the node in case it has some.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -246,4 +246,60 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+            /** @var PropertyInterface $property */
+            $propertyName = $property->getName();
+            if ($propertyName == 'jcr:frozenPrimaryType'
+                || $propertyName == 'jcr:frozenMixinTypes'
+                || $propertyName == 'jcr:frozenUuid'
+            ) {
+                continue;
+            }
+
+            $nodeProperty = $node->getProperty($propertyName);
```
dbu: it could be that the property has been deleted meanwhile. then you would need to create a new property, using the value and the type of the frozen property.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -246,4 +246,60 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+            $onParentVersion = $nodeProperty->getDefinition()->getOnParentVersion();
+            if ($onParentVersion == OnParentVersionAction::COPY || $onParentVersion == OnParentVersionAction::VERSION) {
+                $nodeProperty->setValue($property->getValue());
+            }
+
+            // TODO handle other onParentVersion cases
+        }
```
dbu: would you need to delete properties that where added after the frozen node was created?

dbu: i think you can do some array_intersect_keys magic with getPropertiesValues which gives you an array of all properties of the node. or maybe add a method getPropertyNames to Node, like the getNodeNames method but not part of the NodeInterface.

src/Jackalope/Version/VersionHandler.php
danrot: No, this property is not always set. Removing non-existing properties is a todo a few lines below, it's an extra section in the specification, and also covers other attributes not set.

src/Jackalope/Version/VersionHandler.php
danrot: Ok, so what you are saying is that it would be more save to use `$node->setPropertyValue()`? The problem is the next line, where I have to get the definition from the node property, since the `onParentVerison` attribute is always abort on the frozen node.

src/Jackalope/Version/VersionHandler.php
danrot: When a property was added later the `onParentVersion` attribute influences what is happening. So I am not sure if I just can do some magic with array functions...

src/Jackalope/Version/VersionHandler.php
dbu: yes, but you treat frozenMixinTypes as special case so you need an elseif to remove mixins if the current node has any but frozen has not

src/Jackalope/Version/VersionHandler.php
dbu: then you need to wrap this code into a hasProperty check.

if the property does not exist but frozen, i guess it means that the onparent thing said to copy it, so you can create it with setPropertyValue

src/Jackalope/Version/VersionHandler.php
dbu: the magic would give you the names of properties that exist only on the current node. then you could check their onParentVersion and remove them if needed.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -246,4 +246,60 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+
+        // TODO reset uuid
+
+        foreach ($frozenNode->getProperties() as $property) {
+            /** @var PropertyInterface $property */
+            $propertyName = $property->getName();
+            if ($propertyName == 'jcr:frozenPrimaryType'
+                || $propertyName == 'jcr:frozenMixinTypes'
+                || $propertyName == 'jcr:frozenUuid'
```
dbu: you need to also skip the primary (and possibly mixin) type of the frozen node, and its uuid. so ignore jcr:uuid, jcr:primaryType and jcr:mixinTypes.

maybe you could do an `in_array` to make this a bit more compact.

danrot: Actually this has not to be done, because these properties have a `onParentVersion` value of `IGNORE`, and won't be recovered anyway because of the check later.

dbu: ah okay. fine then.

src/Jackalope/Version/VersionHandler.php
danrot: That's true, will give it a try. But now I always try to write the tests and check in the jackrabbit implementation first :-)

src/Jackalope/Version/VersionHandler.php
danrot: Ok, you are right, will add that.

src/Jackalope/Version/VersionHandler.php
danrot: That might be true, will implement it like that.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -245,5 +247,172 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
 
             $frozenNode->setProperty($propertyName, $property->getValue());
         }
+
+        foreach ($node->getNodes() as $childNode) {
+            /** @var NodeInterface $childNode */
+            // TODO also use the commented lines when getDefinition is implemented
+            //$onParentVersion = $childNode->getDefinition()->getOnParentVersion();
```
danrot: @dbu I would like to also implement the other onParentVersion cases, but the `getDefinition` method of the `Node` is not implemented: https://github.com/jackalope/jackalope/blob/ec1020b8431969708c210c8de88c9a36fd78151f/src/Jackalope/Node.php#L1125

Can this method be almost borrowed from the `Property`?

danrot:
There are still some TODOs in the code. Almost all of them belong the onParentVersion value.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -245,5 +247,172 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+
+            //if ($onParentVersion == OnParentVersionAction::COPY) {
+            // can't use the workspace copy command, because the parent node already has to exist
+            // TODO distinguish between COPY and VERSION onParentVersion
+            $this->copyIntoNode($childNode, $frozenNode);
+            //}
+        }
+    }
+
+    public function restoreItem($removeExisting, $versionPath, $path)
```
dbu: please use `@inheritDoc` when implementing interface methods.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -245,5 +247,172 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+        if ($frozenNode->getPropertyValue('jcr:frozenUuid') != $node->getPropertyValue('jcr:uuid')) {
+            throw new RepositoryException(sprintf(
+                'The frozen node uuid "%s" does not match the actual node uuid "%s". This should never happen.',
+                $frozenNode->getPropertyValue('jcr:frozenUuid'),
+                $node->getPropertyValue('jcr:uuid'))
+            );
+        }
+
+        // TODO reset primary type once the primary type can be changed
+        // also see (https://github.com/jackalope/jackalope/issues/247)
```
dbu: imho we can ignore this for now. as long as the primary type can't be changed, there is also no need to restore. lets keep the comment here though, it would make sense to eventually support this feature.

src/Jackalope/Version/VersionHandler.php
dbu: but the frozen fields do not have that IGNORE?

src/Jackalope/Version/VersionHandler.php
```diff
@@ -245,5 +247,172 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+                || $propertyName == 'jcr:frozenMixinTypes'
+                || $propertyName == 'jcr:frozenUuid'
+            ) {
+                continue;
+            }
+
+            if ($node->hasProperty($propertyName)) {
+                $nodeProperty = $node->getProperty($propertyName);
+                $onParentVersion = $nodeProperty->getDefinition()->getOnParentVersion();
+                if ($onParentVersion == OnParentVersionAction::COPY || $onParentVersion == OnParentVersionAction::VERSION) {
```
dbu: i think you should check this on the properties of the frozen node and just integrate this into the check above and `continue` if the property is not to be copied.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -245,5 +247,172 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+                    $nodeProperty->setValue($property->getValue());
+                }
+            } else {
+                $node->setProperty($propertyName, $property->getValue());
```
dbu: then all that would be left of this if-else should be this line.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -245,5 +247,172 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+            }
+
+            // TODO handle different onParentVersion values
+
+            $childNodePath = $node->getPath() . '/' . $frozenChildNode->getName();
+            if ($this->session->nodeExists($childNodePath)) {
+                $this->session->removeItem($childNodePath);
+            }
+
+            $this->restoreFromNode($node, $frozenChildNode);
```
dbu: why do we first delete the child and then recreate it? should we not just copy the values over, as we do with the node itself?

src/Jackalope/Version/VersionHandler.php
```diff
@@ -245,5 +247,172 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+        foreach ($sourceNode->getNodes() as $childNode) {
+            $this->copyIntoNode($childNode, $copiedNode);
+        }
+    }
+
+    /**
+     * @param NodeInterface $node
+     * @param NodeInterface $frozenNode
+     */
+    private function restoreFromNode(NodeInterface $node, NodeInterface $frozenNode)
```
dbu: that should be $parentNode and $frozenChildNode i think, to make it obvious. (if we need this, see https://github.com/jackalope/jackalope/pull/289/files#r35429501)

src/Jackalope/Version/VersionHandler.php
dbu: yes, i generalized that and created https://github.com/jackalope/jackalope/pull/292

please try if that works for you and give feedback. once its merged, you could rebase the versioning jackalope branch onto master and this PR onto your versioning branch, then you have the feature everywhere.

src/Jackalope/Version/VersionHandler.php
dbu: it should not matter whether the property being restored is currently existing on the node or not.

dbu:
sorry for the long silence. i mostly commented on some details. can you look at those?

for the TODOs i would prefer NotImplementedExceptions. its fine when not all onParentVersion behaviours are implemented in this PR, but throw exceptions when something should happen but does not. ideally explaining what should happen or at least identifying the spec section that is relevant.

it looks to me like it should be possible to quickly wrap this PR up when just throwing exceptions in those cases.

src/Jackalope/Version/VersionHandler.php
danrot: If I stand right, this is the way how the specification describes this. The entire subtree has to be restored as it is in the versioned history. If we just copy over the values the properties being created since the node has been versioned would also continue existing.

danrot:
I guess the tests won't work if the OPVs are not implemented... Hopefully I can have a detailed look at your comment next week.

src/Jackalope/Version/VersionHandler.php
dbu: ah, the children of the previous have to be deleted?

danrot: What do you mean be the previous?

dbu: when i have node A with child B that has again a child C: i restore A' which has B' with a child D': you update the properties of A to those of A', delete B (and implicitly C) and create B' and recursively D'. is that the idea? what if A has other children that do not exist in the version to restore? those will be deleted?

danrot: Yeah, that's the idea, and if there are other nodes they will be deleted. That's the recovering of the entire subgraph from the specification as I understand it.

dbu: seems right to me.

dbu:
ping @danrot

danrot:
Sry, I was on holiday, will have a look at this again soonish.

dbu:
can you rebase this one on the upstream versioning branch please? it is currently in conflict with it.

src/Jackalope/Version/VersionHandler.php
dbu: this is now available in master of jackalope/jackalope

danrot:
@dbu I have rebased to the versioning branch, is that one up to date? Hope I did not crash anything while solving the conflicts...

dbu:
@danrot can you adress the code style comments i made and ping me again? i think this is almost ready to merge.

src/Jackalope/Version/VersionHandler.php
danrot: I don't think this is possible, because the properties on the frozen nodes have different onParentVersion values, IIRC.

danrot:
@dbu Fixed the comments, I think the only thing left now is the correct implementation of the `onParentVersion` attributes, right? Then the entire feature should be ready to merge, or am I missing something?

dbu:
yep, sounds right. can you do that?

danrot:
Yeah, but I can't promise you a due date... should I do that on this PR, or create a new one?

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +236,203 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+
+            switch ($onParentVersion) {
+                case OnParentVersionAction::COPY:
+                case OnParentVersionAction::VERSION:
+                case OnParentVersionAction::ABORT:
+                    $property->remove();
+                    break;
+                case OnParentVersionAction::INITIALIZE:
+                case OnParentVersionAction::COMPUTE: // implementation-specific, decided to handle as INITIALIZE
+                    // TODO how to get default value of property?
```
danrot: I hope this one is ok: When there is a OPV of initialize, the default value of the property should be restored. Compute is implementation specific, so I decided to just do the same as for initialize.

The question that remains is how to get this default value, I couldn't find a method on the definition.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +236,203 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+            if ($this->session->nodeExists($childNodePath)) {
+                $this->session->removeItem($childNodePath);
+            }
+
+            switch ($onParentVersion) {
+                case OnParentVersionAction::COPY:
+                    $this->restoreFromNode($node, $frozenChildNode);
+                    break;
+                case OnParentVersionAction::VERSION:
+
```
danrot: The [specification](http://www.day.com/specs/jcr/2.0/15_Versioning.html#15.7.5 Chained Versions on Restore) leaves a lot of room for implementation specific details here. All it says is that a version label can be used to determine the version, if one is passed along with the restore call. What happens if this is not the case, is not part of the specification. How should we handle this?

src/Jackalope/Version/VersionHandler.php
dbu: afaik the default value is in the property definition. and usually it would be null...

src/Jackalope/Version/VersionHandler.php
dbu: if there is another version of the node currently valid in the workspace, the spec is clear. (depends on removeExisting). otherwise there seems to be a restoreByLabel, which would mean we try to find one with the same label in the history. i guess you meant 

"by a method call that does not specify a label then the workspace in which the restore is being performed will determine which particular version of C will be restored. This determination depends on the configuration of the workspace and is outside the scope of this specification."

i would suggest we just do the easiest thing in this case. can we find what version of the child was active at the point this version was created? if so, we should restore that, otherwise we should restore the last version i guess.

dbu:
btw, did you see that label support is being improved? https://github.com/jackalope/jackalope/pull/302 this will be implemented for jackalope-jackrabbit in master, but could also affect the dbal implementation.

danrot:
@dbu I will consider the new label support, and see if it changes anything for me.

However, I wanted to continue today, but I am having the same issue now as in a different PR. When I am trying to run the tests I am getting loads of these error:

```
DOMDocument::loadXML(): attributes construct error in Entity, line: 1
```

According to `php -i` I still have the libxml version which should not be problematic:

```
libxml

libXML support => active
libXML Compiled Version => 2.9.2
libXML Loaded Version => 20902
libXML streams => enabled
```

Do you still have no idea what could be causing this?

danrot:
@dbu Maybe I should try to getting my branches up to date first... Can you maybe update the `versioning` branches my PRs in jackalope/jackalope and jackalope/doctrine-dbal are targeting?

danrot:
@dbu Found out what the real issue is... The `props`-field of the root node contains this:
```xml
<?xml version="1.0" encoding="UTF-8"?><sv:node xmlns:crx="http://www.day.com/crx/1.0"xmlns:lx="http://flux-cms.org/2.0"xmlns:test="http://liip.to/jackalope"xmlns:mix="http://www.jcp.org/jcr/mix/1.0"xmlns:sling="http://sling.apache.org/jcr/sling/1.0"xmlns:nt="http://www.jcp.org/jcr/nt/1.0"xmlns:fn_old="http://www.w3.org/2004/10/xpath-functions"xmlns:fn="http://www.w3.org/2005/xpath-functions"xmlns:vlt="http://www.day.com/jcr/vault/1.0"xmlns:xs="http://www.w3.org/2001/XMLSchema"xmlns:new_prefix="http://a_new_namespace"xmlns:jcr="http://www.jcp.org/jcr/1.0"xmlns:sv="http://www.jcp.org/jcr/sv/1.0"xmlns:rep="internal" />
```
The point is that the spaces between the attributes are missing. I am getting the same behavior on the master branch. Any ideas?

dbu:
is that in the phpcr-api-tests repo? if you add spaces, does that fix the issue? if so, please do a PR. i supposed it used to be multiline, maybe that somehow got destroyed. (or are you on windows and we have some line ending troubles?)

> Can you maybe update the versioning branches my PRs in jackalope/jackalope and jackalope/doctrine-dbal are targeting?

do you mean rebase them on master?

danrot:
@dbu I found it, was in the jackalope-doctrine-dbal repo, made [another PR](https://github.com/jackalope/jackalope-doctrine-dbal/pull/310) for that. And yes, I mean rebase to master.

dbu:
i updated the versioning branches of jackalope and
jackalope-doctrine-dbal. hope it all went fine. (i merged master into
them, rather than rebasing, as i was afraid of creating a mess with
history with your pull requests and like this conflicts are solved once
and for all)

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +236,203 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+        // handle child nodes present on the frozen node
+        foreach ($frozenNode->getNodes() as $frozenChildNode) {
+            /** @var NodeInterface $frozenChildNode */
+            if (!$removeExisting) {
+                // TODO check occurence of node in repository
+                throw new NotImplementedException('Check for $removeExisting not implemented yet');
+            }
+
+            // TODO handle different onParentVersion values
```
danrot: @dbu I still have to get here the `onParentVersion` attribute of the `$frozenChildNode`. The `$frozenChildNode` says correctly that it is a `nt:unstructure` (at least that's correct for the `VersionManagerTest::testRestoreWithChildren` test, I am currently investigating), but `$frozenChildNode->getDefinition()` returns the definition for `nt:frozenNode`.

It looks to me like this is caused by the algorithm used in `getDefinition`, it takes the parent into account. Looks to me like this doesn't work for frozen nodes.

From what I've seen until now the `onParentVersion` attribute also depends on the parent, so that's probably the reason `getDefinition` makes use of the parent. So what I probably would need is to set the parent on my own. Any elegant way to do so?

dbu: does the spec say anything about node types of children of that frozenNode? maybe the whole subtree should be frozen? or the children would retain their original node type.

either way, getNodeType and getDefinition have to be consistent, i am surprised this is not the case. smells like we have code duplication between getDefinition and getNodeType... do you have the java jackrabbit client running? then you could try to recreate the situation and see what jackrabbit is doing. either the children just keep their node types, or they are frozen too and their original node type is in a separate property, just like for the main frozen node. (the more i think about this, the more sense it would make to freeze the whole subtree. otherwise you could change a child of a frozen node which makes no sense)

danrot: You were right, the children have also the node type `nt:frozenNode`... But that doesn't make my issue easier to solve. I think that means that the logic from `getDefinition` has to be duplicated in some way, because the one existing does not work with all the frozen* properties, or am I missing something?

dbu: it probably means that when restoring, you need to walk the subtree and restore each of the nodes. i don't have in my mind how getDefinition comes into that here? do you need the definition to restore the node? isn't the frozen primary type and the (frozen?) mixins enough? then you could create the child node with  those.

danrot: You are right again :-) The specification is just written a bit strange IMO. [Here](http://www.day.com/specs/jcr/2.0/15_Versioning.html#15.7.6 Restoring Child Nodes) it says that nodes are only copied over if the OPV of the node is COPY or VERSION, but later it says that there will never be a different value on these nodes. So there is no need to check this.

It would make things even easier if the subtree of this node is not frozen at all, since it would be a simple copy command then. But I guess this is still required, because it might be possible that the child of the frozen node is changed by some userland code otherwise, right?

dbu: yes, i think the whole subtree must be frozen to avoid accidentally overwriting. if you want to be sure, best reproduce the situation with jackalope-jackrabbit and check the node types of the children. but afaik a read-only can not be enforced by a parent node so all sub nodes need to be transformed to frozen and back to not frozen.

ahilles107:
Any news here?

danrot:
I am sry, currently I am very busy... IIRC it's not too much work to finish this, but I think I won't find time to finish it before october 😞

dbu:
@ahilles107 would you be interested in trying to take over? if you want to, i can take this branch over into the jackalope repository so you can do pull requests against it.

danrot:
Good News: Since I've been implementing the versioning feature for Sulu, I will now also get some time to work on this, so that we can use the versioning feature also with jackalope-doctrine-dbal :-) Hope I can finish it...

danrot:
@dbu Is there anything new that should be rebased into the `versioning` branch?

dbu:
there is one conflict with master in the versioning branch. but i guess its better to wrap this PR up first before we rebase. otherwise we might get a huge mess with the old versions of the commits that are here in your restore branch.

src/Jackalope/Version/VersionHandler.php
dbu: seems that for jackrabbit they also don't have ignore. so this code above is correct

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +236,196 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+                // TODO check occurence of node in repository
+                throw new NotImplementedException('Check for $removeExisting not implemented yet');
+            }
+
+            $childNodePath = $node->getPath() . '/' . $frozenChildNode->getName();
+            if ($this->session->nodeExists($childNodePath)) {
+                $this->session->removeItem($childNodePath);
+            }
+
+            // TODO handle onParentVersion?
```
danrot: I rewrote the code, but leave the comment here for now. The `onParentVersion` has to be handled in some way, but maybe this information can be inferred in a different way. The only possible OPVs here is `COPY` and `VERSION`, and when it is `VERSION` the resulting node has a different node type, if I understand the [specification](https://docs.adobe.com/content/docs/en/spec/jcr/2.0/3_Repository_Model.html) on 3.13.9 correct:

    Under full versioning, if C is mix:versionable, then a special nt:versionedChild node with a reference to the version history of C is substituted in place of C as a child of the frozen node.

danrot:
@dbu We already said that the children of the frozen nodes must also be frozen nodes, but that is not that easy. Should that be done in a separate PR, or should I fix it in here?

dbu:
lets try to keep this step as small as possible - its big enough already. so if we don't create the children as frozen nodes yet, lets postpone that to a separate PR and do both in a later PR

src/Jackalope/Version/VersionHandler.php
dbu: maybe look for the code around creating a node or search the codebase for the specific property that defines the default value?

src/Jackalope/Version/VersionHandler.php
dbu: sounds right and would need special handling on restore.

danrot:
Ok, so the following steps won't be addressed in this PR:

* restore primary type (we said we won't do that as long as the primary type can't change)
* handle identifier collisions
* handle remove existing

And the following should be fixed before this is being merged:

* handle OPVs when traversing child nodes on current node
* restore the default value on an OPV of `INITIALIZE` or `COMPUTE` (I think that should go into this PR)

src/Jackalope/Version/VersionHandler.php
dbu: check if you always need the exception to be thrown or if the method should return false and we check the return value to know what to do.

erm, actually i guess its the whole isAutocreated part, right? and that only makes sense if the property actually is auto created? we could move the whole switch statement into he node type then.

danrot: I don't think so, because the values for the other properties, which have a special handling in the switch block are not handled by the version at all.

I also dont think it's about the `isAutoCreated` part, since the spec doesn't even mention it:

> If P has an OPV of INITIALIZE then, if N/P has a default value (either defined in the node type of N or implementation-defined) its value is changed to that default value. If N/P has no default value then it is left unchanged.

I think it might also be possible that a property is not auto-created but still have a default value, isn't it? So I think I would go with your first suggestion (the internal  `determineDefaultValue` method on the `PropertyDefinition`).

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +236,199 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+
+            $childNodePath = $node->getPath() . '/' . $frozenChildNode->getName();
+            if ($this->session->nodeExists($childNodePath)) {
+                $this->session->removeItem($childNodePath);
+            }
+
+            if (!$frozenChildNode->isNodeType('nt:versionedChild')) {
+                $this->restoreFromNode($node, $frozenChildNode);
+            } else {
+                // TODO handle restoring from frozenChildNode with OPV of VERSION
```
danrot: I think the check is the correct one, but I would not implement restoring from a child with an OPV of `VERSION` at the moment, because it would make the PR too big.

@dbu Another question is if this must be implemented for the entire Versioning stuff to be merged, or if just throwing a `NotImplementedException` here would be enough.

danrot: I can't imagine that a lot of system want to rely on that anyway, because the specification is not really clear about what happens here:

> Under full versioning each child node C of N where C has an OPV of VERSION and C is versionable, is represented in F not as a copy of N/C but as special node of type nt:versionedChild containing a reference to the version history of C. On restore, N/C in the workspace is replaced by a version of C. The determination of which version of C to use is implementation-dependent (see §15.7.5 Chained Versions on Restore).

(See 15.7.6 on https://docs.adobe.com/content/docs/en/spec/jcr/2.0/15_Versioning.html)

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +236,199 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+        }
+
+        // handle child nodes present on the node but not the frozen node
+        foreach ($node->getNodes() as $childNode) {
+            /** @var NodeInterface $childNode */
+            if ($frozenNode->hasNode($childNode->getName())) {
+                continue;
+            }
+
+            // TODO That's the correct behavior for the OPVs COPY, VERSION and ABORT, handle others as well
```
danrot: I am not a 100% sure if there is anything needed. The specification at https://docs.adobe.com/content/docs/en/spec/jcr/2.0/15_Versioning.html chapter 15.7.6 says this:

> * If C has an OPV of COPY, VERSION or ABORT then N/C is removed. Note that while a node with a child item of OPV ABORT cannot be versioned, it is legal for a previously versioned node to have such a child item added to it and then for it to be restored to the state that it had before that item was added, as this step indicates.
> * If C has an OPV of IGNORE then no change is made to N/C.
> * If C has an OPV of INITIALIZE then N/C is re-initialized as if it were newly created, as defined in its node type.
> * If C has an OPV of COMPUTE then N/C may be re-initialized according to an implementation-specific mechanism.

But when I check 3.13.9 on https://docs.adobe.com/content/docs/en/spec/jcr/2.0/3_Repository_Model.html it only handles the OPV of COPY and VERSION, so checking this value here is probably not necessary, am I right?

src/Jackalope/Version/VersionHandler.php
dbu: right, makes sense to me. i did not check the spec, but your reasoning makes sense to me.

src/Jackalope/Version/VersionHandler.php
dbu: i am ok with leaving parts out, as long as we throw the NotImplementedException and document what we support or not.

src/Jackalope/Version/VersionHandler.php
dbu: does 15.7.6 not apply here? i would think that IGNORE would mean that we don't do the delete that is below. initialize/compute should delete it and then create a new one with default values. why do you think that would not be needed?

src/Jackalope/Version/VersionHandler.php
danrot: Ok, then I'll implement it like that. But I have another question: I would also like to write a test for that, and I would have to add my own node type for this. Hasn't there something like this be done already? I don't know where in the tests I can create my own node-types...

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +237,210 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+            $onParentVersion = $childNode->getDefinition()->getOnParentVersion();
+
+            switch ($onParentVersion) {
+                case OnParentVersionAction::COPY:
+                case OnParentVersionAction::VERSION:
+                case OnParentVersionAction::ABORT:
+                    $childNode->remove();
+                    break;
+                case OnParentVersionAction::INITIALIZE:
+                    // TODO reinitialize node as defined in its node type
```
danrot: I think that's the only thing left before we can merge this. This behavior here should be similar to the one for properties. The code autocreating the child nodes is currently in the [NodeProcessor](https://github.com/jackalope/jackalope/blob/5f59291a191a29fab16c124ffe725fbe3976f823/src/Jackalope/NodeType/NodeProcessor.php#L158). Should I also extract that to a `determineDefaultValue` to the `NodeDefinition`?

src/Jackalope/Version/VersionHandler.php
danrot: Sry, you were right, I mixed that up with a frozen node, but this is not what is handled here.

src/Jackalope/Version/VersionHandler.php
dbu: is it as easy as for a property? what is the "default value" of a node? a node is not just a value.

src/Jackalope/Version/VersionHandler.php
```diff
@@ -235,15 +237,200 @@ private function createFrozenNode(NodeInterface $versionNode, NodeInterface $nod
...
+            $propertyDefinition = $property->getDefinition();
+            $onParentVersion = $propertyDefinition->getOnParentVersion();
+
+            switch ($onParentVersion) {
+                case OnParentVersionAction::COPY:
+                case OnParentVersionAction::VERSION:
+                case OnParentVersionAction::ABORT:
+                    $property->remove();
+                    break;
+                case OnParentVersionAction::INITIALIZE:
```
dbu: implementation-specific means we are free to do whatever the easiest solution is :-)

dbu:
there is NodeTypeTest in phpcr-api-tests. does that help as an
inspiration? if its a generic feature you want to test, we should have
it in phpcr-api-tests. if its about our implementation specific
behaviour, jackalope would be the right place i guess. you can rely on
the phpcr-api-tests infrastructure if you need, i think

src/Jackalope/Version/VersionHandler.php
danrot: Yeah, if I look at the code of the `NodeProcessor` I'd say it's just about adding a node and apply some value to the required properties. Was thinking if it would be enough to just add the node with the primary NodeType and let the system do the rest. Not sure if that will crash something, but then the `NodeProcessor` also wouldn't work. I guess I just give it a try.

PR #291, opened 2015-07-22
added styleci

lsmith77:


src/Jackalope/Lock/LockManager.php
```diff
@@ -260,7 +258,8 @@ public function createLockInfo()
      */
     public function logout()
     {
-        foreach ($this->locks as $path => $lock) {;
+        foreach ($this->locks as $path => $lock) {
+            ;
```
lsmith77: @GrahamCampbell is there a fixer for this as well?

GrahamCampbell: Not sure there is. I've never seen anyone do that with a semicolon before.

dbu: don't you eliminate superflous `;` in general? it would probably make sense even though its not strictly code style but more sort of an error.

GrahamCampbell: We only remove duplicate semicolons atm.

.styleci.yml
```diff
@@ -0,0 +1,8 @@
+preset: psr2
+
+enabled:
+  - long_array_syntax
+  - duplicate_semicolon
+
+disabled:
+  - short_array_syntax
```
GrahamCampbell: You DON'T need this.

GrahamCampbell: This fixer is already disabled by default.

src/Jackalope/Lock/LockManager.php
dbu: fixing that manually for now

PR #295, opened 2015-08-15
when setting an existing property to an empty array, do not try to re…

dbu:
…determine the type to not lose it


lsmith77:
looks good .. I guess this can be merged .. ?

PR #296, opened 2015-08-16
[WIP] Versioning

dbu:
Main PR that will add the wrapped up work of @danrot to master.


mortis0:
Hello, in case one day this MR is picked up, there is a typo in VersionHandler.php line 224 "$frozenNode->hasProperty" should be "$node->hasProperty" and on line 225 "$node->getProperty('jcr:mixinTypes')" should be "$node->getProperty('jcr:mixinTypes')->getValue()"
Minor that, the versionning works fine for me with an XtraDB 8, thanks

PR #297, opened 2015-10-08
ensure that if we fetch the identifier that we have a uuid if the node is referenceable

lsmith77:
help fix https://github.com/jackalope/jackalope-jackrabbit/pull/117

### DIFF
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -21,6 +21,7 @@ use PHPCR\InvalidItemStateException;
 use PHPCR\ItemExistsException;
 use PHPCR\Util\PathHelper;
 use PHPCR\Util\NodeHelper;
+use PHPCR\Util\UUIDHelper;
 use Jackalope\Factory;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Jackalope/Node.php
```diff
@@ -819,6 +825,42 @@ public function getPrimaryItem()
...
+     */
+    public function setUuidGenerator(\Closure $generator)
+    {
+        $this->uuidGenerator = $generator;
+    }
+
+    /**
+     * @return callable a uuid generator function.
+     */
+    protected function getUuidGenerator()
```
dbu: why do we need the additional flexibility with the uuid generator? and is this the best way we can do it? and now that https://github.com/phpcr/phpcr-utils/pull/159 is merged, do we need this?

i would see this more as a 2.0 topic, to make the UuidHelper non-static but an interface that we inject or get from the factory or however we build these things in 2.0.

would it be enough to drop the special case in the ValueConverter and leave this change out?

src/Jackalope/Node.php
```diff
@@ -827,7 +869,11 @@ public function getIdentifier()
     {
         $this->checkState();
 
-        if (isset($this->properties['jcr:uuid'])) {
+        if ($this->isNodeType('mix:referenceable')) {
+            if (empty($this->properties['jcr:uuid'])) {
+                $this->setProperty('jcr:uuid', $this->generateUuid());
+            }
```
dbu: :+1:

src/Jackalope/Node.php
lsmith77: true. i guess we do not need this extension point anymore

lsmith77: done

PR #299, opened 2015-11-04
in outer joins, columns can be null

dbu:
found while investigating the problem @sarcher reported in https://github.com/doctrine/phpcr-odm/pull/667

### DIFF
--- a/src/Jackalope/Query/Row.php
+++ b/src/Jackalope/Query/Row.php
@@ -179,6 +179,12 @@ class Row implements Iterator, RowInterface
      */
     public function getNode($selectorName = null)
     {
+        $path = $this->getPath($selectorName);
+        if (!$path) {
+            // handle outer joins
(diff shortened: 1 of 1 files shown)


src/Jackalope/Query/Row.php
```diff
@@ -179,6 +179,12 @@ public function getValue($columnName)
      */
     public function getNode($selectorName = null)
     {
+        $path = $this->getPath($selectorName);
+        if (!$path) {
+            // handle outer joins
+            return null;
```
dbu: this is the behaviour in jackrabbit.

there is a slight risk for a BC break, if the user relied on the exception being thrown in this case.

dbu:
funnily, the Row code that parses the result is already working correctly (jackrabbit returns all the expected columns with null values). the only problem was isset over array_key_exists

PR #302, opened 2015-11-18
added version label support for the usage with jackalope-jackrabbit

markussr:


src/Jackalope/Version/VersionHistory.php
```diff
@@ -214,9 +230,45 @@ public function hasVersionLabel($label, $version = null)
...
+
+        if ($version === null) {
+            return array_keys($this->versionLabels);
+        } else {
+            $versions = $this->getAllVersions();
+            $versionIsInHistory = false;
+
+            foreach ($versions as $versionCheck) {
+                /* @var VersionInterface $versionCheck */
+                if ($versionCheck->getIdentifier() == $version->getIdentifier()) {
```
dbu: was going to write that $version is a string but you are right. we missed to put the typehint into the interface - we can't change that without BC break now, unfortunately.

src/Jackalope/Version/VersionHistory.php
```diff
@@ -214,9 +230,45 @@ public function hasVersionLabel($label, $version = null)
...
+                    $versionIsInHistory = true;
+                }
+            }
+
+            if (!$versionIsInHistory) {
+                throw new VersionException('version not in history');
```
dbu: please do something like `sprintf('Version %s not found in history of %s', $version->getIdentifier(), $this->getPath())`

src/Jackalope/Version/VersionHistory.php
```diff
@@ -178,9 +188,11 @@ public function getVersionByLabel($label)
...
     public function addVersionLabel($versionName, $label, $moveLabel)
     {
-        // @codeCoverageIgnoreStart
-        throw new NotImplementedException();
-        // @codeCoverageIgnoreEnd
+        $version = $this->getVersion($versionName);
+        $path = $version->getPath();
+
+        $this->objectManager->addLabel($path, $label, $moveLabel);
+        $this->versionLabels[$label] = $version;
```
dbu: what if i did not check for versions before? then versionLabels is not initialized, right?

src/Jackalope/Version/VersionHistory.php
```diff
@@ -166,9 +174,11 @@ public function getVersion($versionName)
      */
     public function getVersionByLabel($label)
     {
-        // @codeCoverageIgnoreStart
-        throw new NotImplementedException();
-        // @codeCoverageIgnoreEnd
+        if ($this->hasVersionLabel($label)) {
+            return $this->versionLabels[$label];
+        } else {
```
dbu: no need for an else, we already returned above. just throwing the exception is enough. or maybe invert, if (!$this->hasVersionLabel...) throw. then return the version.

src/Jackalope/ObjectManager.php
```diff
@@ -970,6 +970,18 @@ public function checkout($absPath)
...
     }
 
+
+    public function addLabel($path, $label, $move)
+    {
+        $this->transport->addLabel($path, $label, $move);
+    }
+
+
+    public function removeLabel($path, $label)
```
dbu: these methods are missing from the VersioningInterface for the transport, right?

adding the methods is a BC break, but only jackrabbit so far implements versions. lets update the branch alias in composer.json to 1.3 and add the methods

dbu: and please add a short phpdoc comment

dbu:
very cool, thanks! technically it looks correct to me, i added a few comments.

src/Jackalope/Version/VersionHistory.php
markussr: We altered the interface VersioningInterface for the transport client.
Maybe we can add the typehint here too?

src/Jackalope/ObjectManager.php
markussr: I've added the methods to the interface and modified the composer.json
But I'm not really sure, if the composer.json is correct. Is it?

src/Jackalope/Version/VersionHistory.php
markussr: The inversion looks nice. I changed the code

src/Jackalope/Version/VersionHistory.php
markussr: Thats right. So I added an init method, that's called in the addVersionLabel and in getVersionLabels

src/Jackalope/Version/VersionHistory.php
markussr: Ok, now an exception with some more information :-)

src/Jackalope/Version/VersionHistory.php
dbu: no, that interface is PHPCR, not the transport client. see https://github.com/phpcr/phpcr/pull/88

src/Jackalope/Version/VersionHistory.php
```diff
@@ -214,9 +231,52 @@ public function hasVersionLabel($label, $version = null)
      */
     public function getVersionLabels($version = null)
     {
-        // @codeCoverageIgnoreStart
-        throw new NotImplementedException();
-        // @codeCoverageIgnoreEnd
+        $this->initVersionLabels();
+        if ($version === null) {
+            return array_keys($this->versionLabels);
+        } else {
```
dbu: no need for the else, as you return if the version is null

src/Jackalope/Version/VersionHistory.php
```diff
@@ -214,9 +231,52 @@ public function hasVersionLabel($label, $version = null)
...
+            $versions = $this->getAllVersions();
+            $versionIsInHistory = false;
+
+            foreach ($versions as $versionCheck) {
+                /* @var VersionInterface $versionCheck */
+                if ($versionCheck->getIdentifier() == $version->getIdentifier()) {
+                    $versionIsInHistory = true;
```
dbu: once we found that the version is in the list, we could `break;` to stop the foreach loop.

src/Jackalope/Version/VersionHistory.php
```diff
@@ -214,9 +231,52 @@ public function hasVersionLabel($label, $version = null)
...
+            }
+
+            return $result;
+        }
+    }
+
+    /**
+     * This method fetches all version labels, if the cache array is not initialized yet.
+     */
+    protected function initVersionLabels()
```
dbu: this method should be private.

src/Jackalope/Version/VersionHistory.php
```diff
@@ -214,9 +231,52 @@ public function hasVersionLabel($label, $version = null)
...
+    {
+        if (is_null($this->versionLabels)) {
```
dbu: lets do the opposite:

```
if (!is_null($this->versionLabels)) {
    return;
}
```

dbu:
i added a few more codestyle things, but otherwise this looks good. can you please add a note about the interface change to the CHANGELOG.md file and squash your commits?

src/Jackalope/Version/VersionHistory.php
dbu: did you forget to push? seems the code still looks the same.

src/Jackalope/Version/VersionHistory.php
```diff
@@ -190,9 +203,13 @@ public function addVersionLabel($versionName, $label, $moveLabel)
...
     {
-        // @codeCoverageIgnoreStart
-        throw new NotImplementedException();
-        // @codeCoverageIgnoreEnd
+        if ($this->hasVersionLabel($label)) {
+            $version = $this->versionLabels[$label];
+            $this->objectManager->removeVersionLabel($version->getPath(), $label);
+            unset($this->versionLabels[$label]);
+        } else {
+            throw new VersionException("No label '$label'");
```
dbu: lets fail early and move the `if` situation out of the if-else. so `if(!$this->hasVersionLabel..) {throw...}`

src/Jackalope/Version/VersionHistory.php
```diff
@@ -214,9 +231,56 @@ public function hasVersionLabel($label, $version = null)
...
+    {
+        if (!is_null($this->versionLabels)) {
+            return;
+        }
+
+        $this->versionLabels = array();
+        $node = $this->getNode('jcr:versionLabels');
+        foreach ($node->getProperties() as $property) {
+            /* @var Property $property */
+            if ($property->getName()!= "jcr:primaryType") {
```
dbu: should we use PHPCR\Util\NodeHelper::isSystemItem? it will basically check whether the name prefix is `jcr:`.

dbu:
i think we are almost there. i added a few comments. can you please also squash your commits into 1?

src/Jackalope/Version/VersionHistory.php
markussr: ok

src/Jackalope/Version/VersionHistory.php
markussr: ok

Issue #303, opened 2015-11-19
Load properties of nodes lazily


danrot:
In the [Node class](https://github.com/jackalope/jackalope/blob/4d3568c42af4132be31f4341c76ad9a1d17d4953/src/Jackalope/Node.php#L251) is already a comment, that the nodes should be loaded lazily. This is a real performance issue for us now, because we have 7 languages on a single node. If we want to load a single language, still all properties get hydrated.

Our solution would be to move the steps from the `_setProperty` method to the `getProperties` method, where the property gets cached. The `_setProperty` would then only cache the passed arguments in a separate array, which information is used to later lazily instantiate the properties in the `getProperties`.

If this solution would be ok for you, we would also create a PR. Just want to make sure we don't invest time in this solution, without it being merged afterwards.

dbu:
i am not sure how much you can gain by that, but this was the idea, yes. if you want to work on that, great!

you would need to check that we don't have the Node class check bypass the getProperty method anywhere just in case, for example when checking its own type and the like. maybe there are a few you need to immediately load to not change the exception behaviour on inconsistencies.

for the problem at hand, child translation strategy might also be a solution, additionally saving you from even fetching all that data from the database.

danrot:
If I have a look at this [profile](https://blackfire.io/profiles/c6ff19dc-38ea-450a-8730-7590fb4bd482/graph?callname=main()&selected=Jackalope%5CNode%3A%3AparseData&settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes) I guess the performance impact of this would be huge :smiley: 

Child nodes for translation would also be a solution, but I think it's much more work than creating this PR, and this PR would also have a positive impact on other areas.

dbu:
okay. cool, go for it!

danrot:
Just to give you an idea of the progress (the comparisons only contain changes in jackalope):

* [Startpage with many links to other pages](https://blackfire.io/profiles/compare/0fd9cf28-826b-4cc0-9f02-5279eb17f7a2/graph)
* [REST response with many pages](https://blackfire.io/profiles/compare/b8c05ced-ab01-416b-8688-db2456918471/graph)
* [Single page](https://blackfire.io/profiles/compare/15ecfe6b-a3c5-4feb-ae5d-7e500cc7df75/graph)

Looks like it only pays of when loading many pages, but then there are really big improvements.

[...]

PR #304, opened 2015-11-24
implemented lazy loading of properties

danrot:
Fixes https://github.com/jackalope/jackalope/issues/303

Tests are running, but I will have a look at the tests of jackalope-doctrine-dbal and jackalope-jackrabbit as well. Will ping here again, when this is done.

### DIFF
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -60,11 +60,18 @@ class Node extends Item implements IteratorAggregate, NodeInterface
      *
      * all properties are instantiated in the constructor
      *
-     * OPTIMIZE: lazy instantiate property objects, just have local array of values
-     *
      * @var Property[]
      */
     protected $properties = [];
+
+    /**
+     * contains all the values for the properties.
+     *
+     * will be used later for lazy instantiation of properties.
+     *
+     * @var array
+     */
+    protected $propertyValues = [];
 
     /**
      * keep track of properties to be deleted until the save operation was successful.
@@ -209,13 +216,14 @@ class Node extends Item implements IteratorAggregate, NodeInterface
                             if ($update) {
                                 unset($oldProperties[$key]);
                             }
-                            if (isset($this->properties[$key])) {
+                            if (isset($this->propertyValues[$key])) {
                                 // refresh existing binary, this will only happen in update
                                 // only update length
-                                if (! ($keepChanges && $this->properties[$key]->isModified())) {
-                                    $this->properties[$key]->_setLength($value);
-                                    if ($this->properties[$key]->isDirty()) {
-                                        $this->properties[$key]->setClean();
+                                $property = $this->loadProperty($key);
+                                if (! ($keepChanges && $property->isModified())) {
+                                    $property->_setLength($value);
+                                    if ($property->isDirty()) {
+                                        $property->setClean();
                                     }
                                 }
                             } else {
@@ -257,11 +265,13 @@ class Node extends Item implements IteratorAggregate, NodeInterface
                         // type information is exposed as property too,
                         // although there exist more specific methods
                         $this->_setProperty('jcr:primaryType', $value, PropertyType::NAME, true);
+                        $this->loadProperty('jcr:primaryType');
                         break;
                     case 'jcr:mixinTypes':
                         // type information is exposed as property too,
                         // although there exist more specific methods
                         $this->_setProperty($key, $value, PropertyType::NAME, true);
+                        $this->loadProperty($key);
                         break;
 
                     // OPTIMIZE: do not instantiate properties until needed
@@ -631,8 +641,8 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         }
 
         if (is_null($value)) {
-            if (isset($this->properties[$name])) {
-                $this->properties[$name]->remove();
+            if (isset($this->propertyValues[$name])) {
+                $this->loadProperty($name)->remove();
             }
 
             return null;
@@ -644,7 +654,16 @@ class Node extends Item implements IteratorAggregate, NodeInterface
             $this->objectManager->registerUuid($value, $this->getPath());
         }
 
-        return $this->_setProperty($name, $value, $type, false);
+        $this->_setProperty($name, $value, $type, false);
+
+        if (isset($this->properties[$name])) {
+            $property = $this->properties[$name];
+            $property->setValue($value, $type ?: PropertyType::UNDEFINED);
+        } else {
+            $property = $this->loadProperty($name);
+        }
+
+        return $property;
     }
 
     /**
@@ -727,6 +746,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
     public function getProperty($relPath)
     {
         $this->checkState();
+        $this->loadProperty($relPath);
 
         if (false === strpos($relPath, '/')) {
             if (!isset($this->properties[$relPath])) {
@@ -826,13 +846,12 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         $this->checkState();
 
         //OPTIMIZE: lazy iterator?
-        $names = self::filterNames($nameFilter, array_keys($this->properties));
+        $names = self::filterNames($nameFilter, array_keys($this->propertyValues));
         $result = [];
-
         foreach ($names as $name) {
             //we know for sure the properties exist, as they come from the
             // array keys of the array we are accessing
-            $result[$name] = $this->properties[$name];
+            $result[$name] = $this->loadProperty($name);
         }
 
         return new ArrayIterator($result);
@@ -853,10 +872,10 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         $this->checkState();
 
         // OPTIMIZE: do not create properties in constructor, go over array here
-        $names = self::filterNames($nameFilter, array_keys($this->properties));
+        $names = self::filterNames($nameFilter, array_keys($this->propertyValues));
         $result = [];
-
         foreach ($names as $name) {
+            $this->loadProperty($name);
             //we know for sure the properties exist, as they come from the
             // array keys of the array we are accessing
             $type = $this->properties[$name]->getType();
@@ -930,7 +949,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         $this->checkState();
 
         if ($this->isNodeType('mix:referenceable')) {
-            if (empty($this->properties['jcr:uuid'])) {
+            if (empty($this->propertyValues['jcr:uuid'])) {
                 $this->setProperty('jcr:uuid', $this->generateUuid());
             }
 
@@ -1006,7 +1025,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         $this->checkState();
 
         if (false === strpos($relPath, '/')) {
-            return isset($this->properties[$relPath]);
+            return isset($this->propertyValues[$relPath]);
         }
         if (! strlen($relPath) || $relPath[0] === '/') {
             throw new InvalidArgumentException("'$relPath' is not a relative path");
@@ -1036,7 +1055,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
     {
         $this->checkState();
 
-        return count($this->properties) !== 0;
+        return count($this->propertiesValues) !== 0;
     }
 
     /**
@@ -1062,14 +1081,14 @@ class Node extends Item implements IteratorAggregate, NodeInterface
     {
         $this->checkState();
 
-        if (!isset($this->properties['jcr:mixinTypes'])) {
+        if (!isset($this->propertyValues['jcr:mixinTypes'])) {
             return [];
         }
 
         $res = [];
         $ntm = $this->session->getWorkspace()->getNodeTypeManager();
 
-        foreach ($this->properties['jcr:mixinTypes']->getValue() as $type) {
+        foreach ($this->loadProperty('jcr:mixinTypes')->getValue() as $type) {
             $res[] = $ntm->getNodeType($type);
         }
 
@@ -1090,8 +1109,8 @@ class Node extends Item implements IteratorAggregate, NodeInterface
             return true;
         }
         // is it one of the mixin types?
-        if (isset($this->properties['jcr:mixinTypes'])) {
-            if (in_array($nodeTypeName, $this->properties["jcr:mixinTypes"]->getValue())) {
+        if (isset($this->propertyValues['jcr:mixinTypes'])) {
+            if (in_array($nodeTypeName, $this->loadProperty("jcr:mixinTypes")->getValue())) {
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


src/Jackalope/Node.php
```diff
@@ -191,15 +198,16 @@ private function parseData($rawData, $update, $keepChanges = false)
...
+                            if (isset($this->propertyValues[$key])) {
                                 // refresh existing binary, this will only happen in update
                                 // only update length
-                                if (! ($keepChanges && $this->properties[$key]->isModified())) {
-                                    $this->properties[$key]->_setLength($value);
-                                    if ($this->properties[$key]->isDirty()) {
-                                        $this->properties[$key]->setClean();
+                                $property = $this->loadProperty($key);
+                                if (! ($keepChanges && $property->isModified())) {
+                                    $property[$key]->_setLength($value);
```
dbu: i think `[$key]` are wrong here (and the 2 lines below)

src/Jackalope/Node.php
```diff
@@ -302,7 +310,9 @@ private function parseData($rawData, $update, $keepChanges = false)
                 $this->nodes = $nodesInBackend;
                 $this->originalNodesOrder = null;
             }
-            foreach ($oldProperties as $name => $property) {
+            foreach ($oldPropertyValues as $name => $propertyValues) {
+                // TODO needs an update
+                $property = $this->loadProperty($name);
```
dbu: i think you try to load the new property here, where you should load the old data.

src/Jackalope/Node.php
```diff
@@ -773,9 +785,10 @@ public function getPropertiesValues($nameFilter = null, $dereference = true)
         $this->checkState();
 
         // OPTIMIZE: do not create properties in constructor, go over array here
-        $names = self::filterNames($nameFilter, array_keys($this->properties));
+        $names = self::filterNames($nameFilter, array_keys($this->propertyValues));
         $result = array();
         foreach ($names as $name) {
+            $this->loadProperty($name);
```
dbu: i think the idea of getPropertiesValues was that you could get the values without needing to instantiate the properties. do we still need this call to do cleanup?

src/Jackalope/Node.php
```diff
@@ -1452,7 +1467,8 @@ public function unsetProperty($name)
      */
     public function confirmSaved()
     {
-        foreach ($this->properties as $property) {
+        foreach ($this->propertyValues as $name => $propertyValue) {
+            $property = $this->loadProperty($name);
             if ($property->isModified() || $property->isNew()) {
```
dbu: i think you don't need loadProperty here. if it was just loaded, it will be neither new nor modified. you could loop over the already existing property objects for this.

src/Jackalope/Node.php
```diff
@@ -1462,17 +1478,6 @@ public function confirmSaved()
     }
 
     /**
-     * In addition to calling parent method, tell all properties
-     */
-    public function setPath($path, $move = false)
```
dbu: i think you still need this for the properties that are already instantiated.

src/Jackalope/Node.php
```diff
@@ -1564,12 +1575,24 @@ public function getIterator()
...
      */
-    protected function _setProperty($name, $value, $type, $internal)
+    protected function loadProperty($name)
     {
+        if (isset($this->properties[$name])) {
+            return $this->properties[$name];
+        }
+
+        if (!isset($this->propertyValues[$name])) {
+            return;
```
dbu: is this the expected behaviour? this makes the method return a mix of property and null. could we throw the not existing exception here?

src/Jackalope/Node.php
```diff
@@ -1564,12 +1575,24 @@ public function getIterator()
...
+
         if ($name == '' | false !== strpos($name, '/')) {
             throw new InvalidArgumentException("The name '$name' is no valid property name");
         }
 
+        $value = $this->propertyValues[$name]['value'];
+        $type = $this->propertyValues[$name]['type'];
+        $internal = $this->propertyValues[$name]['internal'];
+
         if (!isset($this->properties[$name])) {
```
dbu: you already have this check (or rather, the opposite) above. if we get here the property will not exist yet.

src/Jackalope/Node.php
```diff
@@ -1608,7 +1633,8 @@ protected function _setProperty($name, $value, $type, $internal)
     public function setDirty($keepChanges = false, $targetState = false)
     {
         parent::setDirty($keepChanges, $targetState);
-        foreach ($this->properties as $property) {
+        foreach ($this->propertyValues as $name => $propertyValue) {
+            $property = $this->loadProperty($name);
```
dbu: we only need to do this for already loaded properties. the ones not created yet do not need to be re-created later.

src/Jackalope/Node.php
```diff
@@ -1639,7 +1665,8 @@ public function setChildrenDirty()
     public function setDeleted()
     {
         parent::setDeleted();
-        foreach ($this->properties as $property) {
+        foreach ($this->propertyValues as $name => $propertyValue) {
+            $property = $this->loadProperty($name);
```
dbu: same here, only set existing property objects to deleted. all other objects did not exist so there can be no references to them.

src/Jackalope/Node.php
```diff
@@ -1656,8 +1683,9 @@ public function beginTransaction()
         parent::beginTransaction();
 
         // Notify the children properties
-        foreach ($this->properties as $prop) {
-            $prop->beginTransaction();
+        foreach ($this->propertyValues as $name => $propertyValue) {
+            $property = $this->loadProperty($name);
```
dbu: and again here, non existing properties don't need to know about this, i think. or do they?

danrot:
@dbu I'll have a look at your comments soon, but atm I am struggling with the tests... I am getting 800 errors like this (even when running the tests on master):

```
DOMDocument::loadXML(): attributes construct error in Entity, line: 1
```

Any idea what can cause this?

dbu:
dbal or jackrabbit? maybe the version of libxml?

danrot:
dbal, I've got libxml 2.9.2 (according to the `php -i`), shouldn't that be fine?

src/Jackalope/Node.php
danrot: TBH, I don't really know what's happening here. But looks like you are right, so I can't use the `loadProperties` method here... What would be the easiest way to achieve that? Would it be enough to only use the nodes that have already been instantiated before the call of `parseData`?

src/Jackalope/Node.php
danrot: Since afterwards `$this->properties[$name]` is accessed, I was assuming that the property was already instantiated. And since before this PR all properties have been instantiated eagerly, this could have not been the purpose of this method (or at least it didn't work that way), or am I missing something?

src/Jackalope/Node.php
danrot: Ok, so I just revert these changes?

src/Jackalope/Node.php
danrot: I've understood my implementation in a way that `loadProperty` is called for every property (since all the `_setProperty` calls do not instantiate properties anymore). And I've moved the `setPath` call there, is there a scenario where this would not work?

src/Jackalope/Node.php
danrot: Had to put this here in order to be able to save nodes again... So I guess just throwing another exception would also not work... I'll try to find another solution.

src/Jackalope/Node.php
danrot: So that means I can delete the else path? I am not totally sure about this.

src/Jackalope/Node.php
danrot: I don't really know... Shouldn't a property that is added later have the same transaction state? However, if this is removed, it should also be removed in the `commitTransaction` and `rollbackTransaction` methods, right?

src/Jackalope/Node.php
dbu: sorry i was unclear. we added this method (among other reasons) to provide ourselves a way to *in the future* provide access to the values without needing to instantiate properties.

however, its probably a bit tricky, depending on the type (reference, date, ...) we would need to convert a string value so it could get really quite hairy. we could check if we have the property and if not, create it unless its a string/int/float property in which case we can just use the value as is, wdyt?

src/Jackalope/Node.php
dbu: yes please. maybe add a comment to avoid confusion in the future.

src/Jackalope/Node.php
dbu: at least when we have a move operation, we must propagate that to the already instantiated properties. i think you should just do the same we have now (no need to load not yet loaded properties)

src/Jackalope/Node.php
dbu: hm, maybe the github diff is confusing me here. but i would not see that you can ever reach the else path below. look at line 1580 - if the property is already set, you return it.

src/Jackalope/Node.php
dbu: yes. i guess what you however do need is if you load a property while inside a transaction, pass the transaction status to that property. otherwise the commit/rollback will be a problem on a property that was loaded while inside the transaction.

src/Jackalope/Node.php
dbu: this is part of the refresh operation. i think it was about removing newly created properties again if we refresh and they where not saved. so you could just have an $oldProperties with the already loaded properties and loop over those here. if they are newly created, they certainly are loaded.

src/Jackalope/Node.php
```diff
@@ -213,9 +221,9 @@ private function parseData($rawData, $update, $keepChanges = false)
                     continue;
                 }
 
-                if ($update && array_key_exists($key, $this->properties)) {
-                    unset($oldProperties[$key]);
-                    $prop = $this->properties[$key];
+                if ($update && array_key_exists($key, $this->propertyValues)) {
+                    unset($oldPropertyValues[$key]);
+                    $prop = $this->loadProperty($key);
```
dbu: i think same here, this is only about properties that are already loaded. if they are not loaded, they can't be modified or deleted (or does deleting a property using Node::setValue not lead to the property being loaded?)

dbu:
it should. did the problem occur before your changes as well?

src/Jackalope/Node.php
danrot: How cheap is it to get the type of a property? At this place there's only the name, so I have to load something either way. I am not sure if this causes a notable performance gain, and therefore is worth the duplication of this logic.

src/Jackalope/Node.php
danrot: Yeah, that's what I thought, so I've moved this logic to https://github.com/jackalope/jackalope/pull/304/files#diff-194b8b91be1497a30eb4ca5256bc194bR1623

src/Jackalope/Node.php
danrot: I know, but I am sure that the code in the else path was there for a reason. Maybe there's a caveat in this new strategy? But if you say it should be save to delete I'll do it.

src/Jackalope/Node.php
danrot: Would you also do that in this PR? Or can that be a separate one? (since this one is more about read performance)

src/Jackalope/Node.php
danrot: Makes sense, will change it

danrot:
Yeah, that's the confusing part... However, I don't get the error with jackrabbit, will test there first.

danrot:
I think before you further review it makes sense that I get this completely running first. Might have to change some stuff anyway.

src/Jackalope/Node.php
dbu: assume this:

```php
$n = $session->getNode('/bar');
$p = $n->getProperty('foo');
$session->move('/bar', '/other');
echo $p->getPath();
```

we do still need setPath, but only update the properties that are already loaded

src/Jackalope/Node.php
dbu: the previous code was about _setProperty. that is not the same as loadProperty. not sure if we can eliminate _setProperty totally.

src/Jackalope/Node.php
dbu: i don't think there is much missing. just revert the changes on the transaction methods and pass the transaction status in loadProperty should do the trick. the problem is that it will become hard to see what is going on once refactored this way.

src/Jackalope/Node.php
```diff
@@ -580,14 +588,23 @@ public function setProperty($name, $value, $type = PropertyType::UNDEFINED, $val
...
 
-        return $this->_setProperty($name, $value, $type, false);
+        $this->_setProperty($name, $value, $type, false);
+
+        if (isset($this->properties[$name])) {
+            $property = $this->properties[$name];
+            $property->setValue($value, $type ?: PropertyType::UNDEFINED);
+        } else {
+            $property = $this->loadProperty($name);
+        }
```
danrot: I think I have to find a cleaner solution here... But for now I just try to get the tests with jackalope-jackrabbit running again, and clean up afterwards.

danrot:
@dbu I am having troubles with the `testAddSessionRefresh` test in `CombinedManipulationsTest`. The problem is that I don't know how the property disappears on the call https://github.com/phpcr/phpcr-api-tests/blob/master/tests/Writing/CombinedManipulationsTest.php#L695. I've debugged it, and I realize that the `prop` property is just gone after executing this line: https://github.com/jackalope/jackalope/blob/master/src/Jackalope/Item.php#L613

It looks like a simple comparison, but yet the property seems to be deleted afterwards. How does this work?

danrot:
ping @dbu

akira28:
any news on this? We are hitting a wall importing several thousands pages, and could solve the performance issue we are facing.
@danrot Is your fork usable?

akira28:
full disclosure, we are working on a Sulu project, so you should have a lot of insight on this particular issue :D

dbu:
@danrot are you having a look into this? please rebase the branch. if you are still stuck on the failure, i can have another look at it next week when i am back from canada.

akira28:
thanks @danrot for rebasing 👍

danrot:
No problem. Have seen that the tests are already passing here, but there are still some problems when running the tests in jackalope-jackrabbit. Haven't tested jackalope-doctrine-dbal.

akira28:
@dbu have you had the opportunity to look at it now that's rebased?

dbu:
@danrot can you create a pull request on jackalope-jackrabbit and jackalope-doctrine-dbal that use this branch instead of the normal source (you will need to custom set where the repository is found) so that we can see on travisci how things look?

danrot:
@dbu Created https://github.com/jackalope/jackalope-jackrabbit/pull/140 and https://github.com/jackalope/jackalope-doctrine-dbal/pull/331

danrot:
Actually it doesn't look tooo bad 😃

dbu:
yeah, seems to be mainly a couple of things with versioning and the combined manipulations.

can you go over the code again with a fresh mind and check in each place whether we should check the instantiated properties or the raw values, and make sure we always make sure to instantiate the property if needed? i feel there are some mixups what we use when. that could also explain why values are not updated in some situations.

also, diff calculation in transport should have a way to only look at properties that actually changed instead of creating instantiation of all of them. but that can be a separate pull request that comes later - its also kindof a BC break so we would need to check from transport if the jackalope version is new enough.

danrot:
Sadly I am busy with other stuff currently, can't promise a due date to that 😕

dbu:
ping @danrot

danrot:
Sry, I really don't have any time available for working at this at the moment 😕

PR #306, opened 2016-02-05
improve node type validation reporting

dbu:
during debugging, i realized its really hard to find out what triggers a contraint validation error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
@@ -14,7 +13,7 @@ sudo: false
 
 matrix:
   include:
-    - php: 5.3.3
+    - php: 5.3
       env: PACKAGE_VERSION=low
 
 before_script:
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -558,7 +558,13 @@ class Node extends Item implements IteratorAggregate, NodeInterface
                     }
                 }
                 if (null !== $exception) {
-                    throw $exception;
+                    $types = 'Primary type '.$this->primaryType;
+                    if (isset($this->properties['jcr:mixinTypes'])) {
+                        $types .= ', mixins '.implode(',', $this->getPropertyValue('jcr:mixinTypes', PropertyType::STRING));
+                    }
+                    $msg = sprintf('Can not set property %s on node %s. Node types do not allow for this: %s', $name, $this->path, $types);
+
+                    throw new ConstraintViolationException($msg, 0, $exception);
                 }
             } else {
                 // $value is null for property removal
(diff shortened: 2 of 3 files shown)


danrot:
I am getting an error on [appveyor](https://ci.appveyor.com/project/sulu/sulu/build/1.0.4446#L796), which seems to be caused by this change...

```
PHP Warning:  Uncaught var_export does not handle circular references
C:\projects\sulu\vendor\symfony\phpunit-bridge\DeprecationErrorHandler.php:73
C:\projects\sulu\vendor\jackalope\jackalope\src\Jackalope\NodeType\NodeType.php:253
C:\projects\sulu\vendor\jackalope\jackalope\src\Jackalope\Node.php:551
C:\projects\sulu\src\Sulu\Component\Content\Document\Subscriber\StructureRemoveSubscriber.php:163
C:\projects\sulu\src\Sulu\Component\Content\Document\Subscriber\StructureRemoveSubscriber.php:136
C:\projects\sulu\src\Sulu\Component\Content\Document\Subscriber\StructureRemoveSubscriber.php:117
C:\projects\sulu\src\Sulu\Component\Content\Document\Subscriber\StructureRemoveSubscriber.php:94
C:\projects\sulu\src\Sulu\Component\Content\Document\Subscriber\StructureRemoveSubscriber.php:82
C:\projects\sulu\vendor\symfony\symfony\src\Symfony\Component\EventDispatcher\EventDispatcher.php:174
C:\projects\sulu\vendor\symfony\symfony\src\Symfony\Component\EventDispatcher\EventDispatcher.php:43
C:\projects\sulu\vendor\sulu\document-manag in C:\projects\sulu\vendor\jackalope\jackalope\src\Jackalope\NodeType\NodeType.php on line 253
PHP Fatal error:  Allowed memory size of 3221225472 bytes exhausted (tried to allocate 1572868096 bytes) in C:\projects\sulu\vendor\jackalope\jackalope\src\Jackalope\NodeType\NodeType.php on line 253
```

This only happens on the windows build from appveyor, not on the linux builds on travis. Any ideas? Is this a windows issue? That's probably not caught because there are no windows builds in this repo, right?

dbu:
yeah we don't do windows builds on gitlab-ci. var_export is "only" about reporting information when things go wrong. if you have an idea how to make this less dangerous, i am happy for a pull request. if we reduce the amount of information we can output, i think we have to life with that drawback - better than completely exploding.

danrot:
What do you mean by "better than completely exploding"? It would be better to have the worse exception message but it is not failing like in my case?

And it is also not completely clear to me what this PR does... Which exact information do you need in this place? Hard to make any suggestions without that knowledge...

dbu:
> "better than completely exploding"

avoid to use all memory on a recursion. which not only breaks the process but if it happens on a production could bring the whole server into troubles.

> And it is also not completely clear to me what this PR does.

the idea was to better report what the node type definition allows and what was passed in, when validation failed. i think instead of var_export we could first check if the $value is an array of scalars or something else, and translate that to "nested array" or "array of objects" and only export arrays of scalars.

danrot:
@dbu Created a PR for that: https://github.com/jackalope/jackalope/pull/333

PR #309, opened 2016-03-31
Do not throw exception on checkState if item is deleted / set node clean after save/update.

dantleech:
Set the node as clean after updating the node via. transport during the `save()` operation.

See related:
- Jackalope Doctrine DBAL: https://github.com/jackalope/jackalope-doctrine-dbal/pull/323
- PHPCR API tests: https://github.com/phpcr/phpcr-api-tests/pull/178/files

### DIFF
--- a/src/Jackalope/ObjectManager.php
+++ b/src/Jackalope/ObjectManager.php
@@ -869,6 +869,7 @@ class ObjectManager
                 if ($item->isModified() || $item->isMoved()) {
                     $item->confirmSaved();
                 }
+                $item->setClean();
             }
         }
 
@@ -1121,10 +1122,10 @@ class ObjectManager
 
         /** @var $node Node */
         foreach ($this->objectsByPath['Node'] as $node) {
-            if (! $keepChanges || ! ($node->isDeleted() || $node->isNew())) {
+            if (false === $keepChanges || ! ($node->isDeleted() || $node->isNew())) {
                 // if we keep changes, do not restore a deleted item
                 $this->objectsByUuid[$node->getIdentifier()] = $node->getPath();
-                $node->setDirty($keepChanges);
+                $node->setDirty(false);
             }
         }
     }
@@ -1812,6 +1813,7 @@ class ObjectManager
             if (false !== $uuid) {
                 unset($this->objectsByUuid[$uuid]);
             }
+
             unset($this->objectsByPath['Node'][$absPath]);
             $item->setDeleted();
         }
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -1308,7 +1308,7 @@ class Node extends Item implements IteratorAggregate, NodeInterface
         } catch (ItemNotFoundException $ex) {
 
             // The node was deleted in another session
-            if (! $this->objectManager->purgeDisappearedNode($this->path, $keepChanges)) {
+            if (false === $this->objectManager->purgeDisappearedNode($this->path, $keepChanges)) {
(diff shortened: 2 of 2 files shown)


src/Jackalope/Item.php
```diff
@@ -636,10 +636,6 @@ protected function checkState()
             $this->postDirtyState = -1;
         }
 
-        if ($this->state === self::STATE_DELETED) {
-            throw new InvalidItemStateException('Item '.$this->path.' is deleted');
-        }
```
dantleech: So I guess this implies that `DELETED` is a valid state until you try and manipulate the node.

dbu: the problem with this is that when you manipulate the node, it calls checkState to make sure it can be manipulated. 

looking at the stack trace in https://travis-ci.org/jackalope/jackalope-doctrine-dbal/jobs/119277489 i think the problem is something else. https://github.com/jackalope/jackalope/blob/check_state_delete_no_exception/src/Jackalope/ObjectManager.php#L1124 assumes that a node that is dirty is not deleted (because both are states). but then ->getIdentifier will trigger checkState which detects the node is deleted. i think before the deleted check in the object manager, we should check for dirty and if the node is dirty call refresh($keepChanges). then the state will not be dirty anymore but deleted or new or modified and things work as expected i'd hope.

dbu: if you call refresh repeatedly, things will become very expensive as everything that is dirty is reloaded.

it smells fishy when a node is still dirty after a save() operations, btw. if that is the case there might be some other bug too.

src/Jackalope/ObjectManager.php
```diff
@@ -869,6 +869,7 @@ public function save()
                 if ($item->isModified() || $item->isMoved()) {
                     $item->confirmSaved();
                 }
+                $item->setClean();
```
dantleech: This is the change.

dantleech:
@dbu have updated - now the only change is setting the node to clean after updating it with the transport, seems to fix the issue. When the user later tries to access a property of the (deleted) node they will receive the usual InvalidItemStateException.

src/Jackalope/ObjectManager.php
dbu: i think we did not do this so that auto generated things would be found when reloading (jackrabbit might autocreate some things)

dantleech: Not sure I understand: do you mean if the node contains some auto-created properties? i.e. if it is dirty then the state will be reloaded somehow?

Otherwise do you think this is the correct behavior? If not, any ideas as to what we can do to fix the issue?

dbu: i think we on purpose mark the node dirty when we confirm it its saved. there are autocreated properties in the backend, e.g. the uuid if you did not force creation already.


i wrote about the solution that i expect to be correct in a comment that is now hidden:

looking at the stack trace in https://travis-ci.org/jackalope/jackalope-doctrine-dbal/jobs/119277489 i think the problem is something else. https://github.com/jackalope/jackalope/blob/check_state_delete_no_exception/src/Jackalope/ObjectManager.php#L1124 assumes that a node that is dirty is not deleted (because both are states). but then ->getIdentifier will trigger checkState which detects the node is deleted. i think before the deleted check in the object manager, we should check for dirty and if the node is dirty call refresh($keepChanges). then the state will not be dirty anymore but deleted or new or modified and things work as expected i'd hope.

dbu: @dantleech can we do the solution i propose in my last comment, or does that lead to some other problem?

PR #316, opened 2016-09-22
fixed some phpDocs

vondrasekpavel:


src/Jackalope/Transport/TransportInterface.php
```diff
@@ -322,7 +322,7 @@ public function setNodeTypeManager($nodeTypeManager);
      * @return array with the definitions (see
      *      Jackalope\NodeTypeDefinition::fromArray for what is expected)
```
dbu: can you also remove the (see ...) here? the `@see` below covers that.

src/Jackalope/Transport/TransportInterface.php
```diff
@@ -322,7 +322,7 @@ public function setNodeTypeManager($nodeTypeManager);
...
      *
-     * @see Jackalope\NodeTypeDefinition::fromArray
+     * @see \Jackalope\NodeTypeDefinition::fromArray
```
dbu: please use a `use` statement for this as well, so here it will be `@see NodeTypeDefinition::fromArray`.

src/Jackalope/Transport/TransportInterface.php
vondrasekpavel: Ok, i will fix it :)

dbu:
thanks a lot!

PR #318, opened 2016-09-22
fixed some phpDocs & other details

vondrasekpavel:
Hi,

I fixed some phpDocs & changed some details.

Please check it because I'm groggy and some methods don't have tests :(

### DIFF
--- a/src/Jackalope/Helper.php
+++ b/src/Jackalope/Helper.php
@@ -33,6 +33,6 @@ class Helper
             throw new RepositoryException("Expected attribute $attribute not found on ".$node->getNodePath());
         }
 
-        return !('false' === $node->getAttribute($attribute));
+        return 'false' !== $node->getAttribute($attribute);
     }
 }
--- a/src/Jackalope/Helper.php
+++ b/src/Jackalope/Helper.php
@@ -24,16 +24,15 @@ class Helper
      * @param string     $attribute name to fetch
      *
      * @return bool the value converted to bool
+     *
+     * @throws RepositoryException
      */
     public static function getBoolAttribute(DOMElement $node, $attribute)
     {
         if (! $node->hasAttribute($attribute)) {
             throw new RepositoryException("Expected attribute $attribute not found on ".$node->getNodePath());
         }
-        if ('false' === $node->getAttribute($attribute)) {
-            return false;
-        }
 
-        return true;
+        return !('false' === $node->getAttribute($attribute));
     }
 }
--- a/src/Jackalope/ImportExport/ImportExport.php
+++ b/src/Jackalope/ImportExport/ImportExport.php
@@ -49,7 +49,7 @@ class ImportExport implements ImportUUIDBehaviorInterface
      * @param boolean                    $skipBinary as in exportSystemView
      * @param boolean                    $noRecurse  as in exportSystemView
      *
-     * @see PHPCR\SessionInterface::exportSystemView
+     * @see \PHPCR\SessionInterface::exportSystemView
      */
     public static function exportSystemView(NodeInterface $node, NamespaceRegistryInterface $ns, $stream, $skipBinary, $noRecurse)
     {
@@ -66,7 +66,9 @@ class ImportExport implements ImportUUIDBehaviorInterface
      * @param boolean                    $skipBinary as in exportDocumentView
      * @param boolean                    $noRecurse  as in exportDocumentView
      *
-     * @see PHPCR\SessionInterface::exportDocumentView
+     * @throws RepositoryException
+     *
+     * @see \PHPCR\SessionInterface::exportDocumentView
      */
     public static function exportDocumentView(NodeInterface $node, NamespaceRegistryInterface $ns, $stream, $skipBinary, $noRecurse)
     {
@@ -82,14 +84,17 @@ class ImportExport implements ImportUUIDBehaviorInterface
      * @param string                     $uri          as in importXML
      * @param integer                    $uuidBehavior as in importXML
      *
-     * @see PHPCR\SessionInterface::importXML
+     * @throws \RuntimeException
+     * @throws \Exception
+     *
+     * @see \PHPCR\SessionInterface::importXML
      */
     public static function importXML(NodeInterface $parentNode, NamespaceRegistryInterface $ns, $uri, $uuidBehavior)
     {
         $use_errors = libxml_use_internal_errors(true);
         libxml_clear_errors();
 
-        if (! file_exists($uri)) {
+        if (!file_exists($uri)) {
             throw new \RuntimeException("File $uri does not exist or is not readable");
         }
 
@@ -106,7 +111,7 @@ class ImportExport implements ImportUUIDBehaviorInterface
 
         $xml->read();
         try {
-            if ('node' == $xml->localName && NamespaceRegistryInterface::NAMESPACE_SV == $xml->namespaceURI) {
+            if ('node' === $xml->localName && NamespaceRegistryInterface::NAMESPACE_SV === $xml->namespaceURI) {
                 // TODO: validate with DTD?
                 self::importSystemView($parentNode, $ns, $xml, $uuidBehavior);
             } else {
@@ -144,10 +149,11 @@ class ImportExport implements ImportUUIDBehaviorInterface
      * prefix if needed.
      *
      * @param string $name A name encoded with escapeXmlName
+     * @param array $namespaceMap
      *
      * @return string the decoded name
      */
-    public static function unescapeXmlName($name, $namespaceMap)
+    public static function unescapeXmlName($name, array $namespaceMap)
     {
         foreach (self::$escaping as $raw => $escaped) {
             // Used a negative look behind to only replace non-escaped escape characters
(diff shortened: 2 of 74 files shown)


src/Jackalope/ObjectManager.php
```diff
@@ -785,6 +801,7 @@ public function registerNodeTypesCnd($cnd, $allowUpdate)
                 $nsRegistry->registerNamespace($prefix, $uri);
             }
 
+            // TODO: Is @return type bool wrong?
```
dbu: yes, the `@return` is wrong. this is called by NodeTypeManager::registerNodeTypesCnd which states its an \Iterator.

rather than put the whole description of that iterator here, we could do `@see NodeTypeManagerInterface::registerNodeTypesCnd`

src/Jackalope/Helper.php
```diff
@@ -24,16 +24,15 @@ class Helper
...
     {
         if (! $node->hasAttribute($attribute)) {
             throw new RepositoryException("Expected attribute $attribute not found on ".$node->getNodePath());
         }
-        if ('false' === $node->getAttribute($attribute)) {
-            return false;
-        }
 
-        return true;
+        return !('false' === $node->getAttribute($attribute));
```
vondrasekpavel: yes, you are right :)

src/Jackalope/ImportExport/ImportExport.php
```diff
@@ -49,7 +49,7 @@ class ImportExport implements ImportUUIDBehaviorInterface
      * @param boolean                    $skipBinary as in exportSystemView
      * @param boolean                    $noRecurse  as in exportSystemView
      *
-     * @see PHPCR\SessionInterface::exportSystemView
+     * @see \PHPCR\SessionInterface::exportSystemView
```
vondrasekpavel: ok

src/Jackalope/ImportExport/ImportExport.php
```diff
@@ -66,7 +66,9 @@ public static function exportSystemView(NodeInterface $node, NamespaceRegistryIn
      * @param boolean                    $skipBinary as in exportDocumentView
      * @param boolean                    $noRecurse  as in exportDocumentView
      *
-     * @see PHPCR\SessionInterface::exportDocumentView
+     * @throws RepositoryException
+     *
+     * @see \PHPCR\SessionInterface::exportDocumentView
```
vondrasekpavel: ok

src/Jackalope/ImportExport/ImportExport.php
```diff
@@ -244,7 +254,7 @@ private static function addNode(NodeInterface $parentNode, $nodename, $type, $pr
             $type = 'nt:unstructured';
         }
 
-        if ('jcr:root' == $nodename && isset($existing) && $existing->getDepth() === 0 && self::IMPORT_UUID_COLLISION_REPLACE_EXISTING == $uuidBehavior) {
+        if ('jcr:root' === $nodename && isset($existing) && self::IMPORT_UUID_COLLISION_REPLACE_EXISTING === $uuidBehavior && $existing->getDepth() === 0) {
```
vondrasekpavel: ok

src/Jackalope/ImportExport/ImportExport.php
```diff
@@ -320,11 +334,12 @@ private static function exportSystemViewRecursive(NodeInterface $node, Namespace
 
         foreach ($node->getProperties() as $name => $property) {
             /** @var $property \PHPCR\PropertyInterface */
-            if ($name == 'jcr:primaryType' || $name == 'jcr:mixinTypes' || $name == 'jcr:uuid') {
+
+            if ($name === 'jcr:primaryType' || $name === 'jcr:mixinTypes' || $name === 'jcr:uuid') {
```
vondrasekpavel: ok

src/Jackalope/Lock/Lock.php
```diff
@@ -90,6 +92,10 @@ public function setIsDeep($isDeep)
      * {@inheritDoc}
      *
      * @api
+     *
+     * @throws NotImplementedException
```
vondrasekpavel: ok, I removed it :)

src/Jackalope/Node.php
```diff
@@ -597,7 +598,7 @@ public function getNode($relPath)
     {
         $this->checkState();
 
-        if (strlen($relPath) == 0 || '/' == $relPath[0]) {
+        if ($relPath === '' || '/' === $relPath[0]) {
```
vondrasekpavel: ok

src/Jackalope/ObjectManager.php
```diff
@@ -779,9 +781,17 @@ protected function pathArrayToPropertiesIterator($propertyPaths)
      *
      * @param  string  $cnd         a string with cnd information
      * @param  boolean $allowUpdate whether to fail if node already exists or to update it
-     * @return bool    true on success
+     * @return bool|\Iterator    true on success or \Iterator over the registered node types if repository is not able to process
```
dbu: i think this is not correct. this method always registers the node types, either directly through transport or by parsing the file and calling the node type manager registerNodeTypes method.

if neither is possible, it throws UnsupportedRepositoryOperationException

so no boolean is ever returned.

vondrasekpavel: Ok, think we must repair interface because bool return type is defined here:

snimek obrazovky 2016-09-23 v 14 46 11

src/Jackalope/Transport/AbstractReadLoggingWrapper.php
```diff
@@ -191,7 +191,7 @@ public function setNodeTypeManager($nodeTypeManager)
     /**
      * {@inheritDoc}
      */
-    public function getNodeTypes($nodeTypes = array())
+    public function getNodeTypes(array $nodeTypes = array())
```
dbu: i had to revert this change to the interface you did in the first PR, as its a BC break for the implementations. you need to remove it here too.

src/Jackalope/Transport/NodeTypeFilterInterface.php
```diff
@@ -30,7 +30,7 @@
      *
      * @see TransportInterface::getNodes
      */
-    public function getNodesFiltered($paths, $typeFilter);
+    public function getNodesFiltered(array $paths, $typeFilter);
```
dbu: this is a BC break, please revert

src/Jackalope/Transport/TransportInterface.php
```diff
@@ -319,12 +324,11 @@ public function setNodeTypeManager($nodeTypeManager);
...
      *
-     * @return array with the definitions (see
-     *      Jackalope\NodeTypeDefinition::fromArray for what is expected)
+     * @return array with the definitions
      *
-     * @see Jackalope\NodeTypeDefinition::fromArray
+     * @see NodeTypeDefinition::fromArray
      */
-    public function getNodeTypes($nodeTypes = array());
+    public function getNodeTypes(array $nodeTypes = array());
```
dbu: this is a BC break. please rebase on the upstream master, where i reverted this.

vondrasekpavel:
Oh, sorry :( 

For now I will revert this change. I will try to fix it later.

src/Jackalope/ImportExport/ImportExport.php
```diff
@@ -464,10 +502,10 @@ private static function importSystemView(NodeInterface $parentNode, NamespaceReg
         if (! $xml->read()) {
             throw new InvalidSerializedDataException('missing information to create node');
         }
-        if ('property' != $xml->localName || NamespaceRegistryInterface::NAMESPACE_SV != $xml->namespaceURI) {
+        if ('property' !== $xml->localName || NamespaceRegistryInterface::NAMESPACE_SV !== $xml->namespaceURI) {
             throw new InvalidSerializedDataException('first child of node must be sv:property for jcr:primaryType. Found {'.$xml->namespaceURI.'}'.$xml->localName.'="'.$xml->value.'"'.$xml->nodeType);
         }
-        if (! $xml->moveToAttributeNs('name', NamespaceRegistryInterface::NAMESPACE_SV) || 'jcr:primaryType' != $xml->value) {
+        if ('jcr:primaryType' !== $xml->value || !$xml->moveToAttributeNs('name', NamespaceRegistryInterface::NAMESPACE_SV)) {
```
dbu: i tested with jackalope-doctrine-dbal, and this line is a regression. the problem is that we call moveToAttributeNs for its side effect of moving. plus, the check for value must happen after we moved to the attribute, not before.

to make this more explicit, we could split this in two like this:

```
if (! $xml->moveToAttributeNs('name', NamespaceRegistryInterface::NAMESPACE_SV)) {
    throw new InvalidSerializedDataException('first child of node must have a sv:name property');
}
if ('jcr:primaryType' !== $xml->value) {
    throw new InvalidSerializedDataException('first child of node must be the sv:property node with a jcr:primaryType. Found {'.$xml->namespaceURI.'}'.$xml->localName.'="'.$xml->value
}
```

do you want to update like this? i think if you change that, this would be ready to merge.

(if you want, you could search this file for " == " - i found some places where we don't use === that you missed.

dbu:
please rebase on master, github sees a conflict.

if you want, you could search this file for " == " - i found some places where we don't use === that you missed.

src/Jackalope/ImportExport/ImportExport.php
vondrasekpavel: ok, I will :)

dbu:
great, tests with jackalope-doctrine-dbal are now green. can you please squash the commits?

dbu:
squashed and merged manually in 5f59291a191a29fab16c124ffe725fbe3976f823

thanks a lot for this cleanup work @vondrasekpavel !

Issue #322, opened 2016-12-21
Access empty arrays fails due to non-existing index


danrot:
When we write an empty array to a field, we cannot access it later in a query anymore, because there is [code accessing the first entry](https://github.com/jackalope/jackalope/blob/5f59291a191a29fab16c124ffe725fbe3976f823/src/Jackalope/Query/Row.php#L165), which fails with a notice undefined offset.

This is happening on jackalope-doctrine-dbal, but not on jackalope-jackrabbit. I guess the empty array should not be written at all, right? Or should there be another check in teh row class?

dbu:
setting a property to null deletes it. but having a multivalue property
with 0 entries does not delete so that should be supported. i think we
should just add a `count($value)` as first check on the line you linked.
can you do a pull request?

danrot:
Fixed in #323

PR #323, opened 2017-01-09
added check for empty arrays when resolving queries

danrot:
This PR adds the check for empty arrays as suggested in https://github.com/jackalope/jackalope/issues/322#issuecomment-268827882

### DIFF
--- a/src/Jackalope/Query/Row.php
+++ b/src/Jackalope/Query/Row.php
@@ -163,7 +163,7 @@ class Row implements Iterator, RowInterface
         // for multi-values boolean/binary values we can't provide a
         // defined result so we return null
         if (is_array($value)) {
-            if (is_scalar($value[0]) && !is_bool($value[0])) {
+            if (count($value) && is_scalar($value[0]) && !is_bool($value[0])) {
                 $value = implode(' ', $value);
             } else {
                 $value = null;


danrot:
@dbu There are no tests for this class, are there?

src/Jackalope/Query/Row.php
```diff
@@ -162,7 +162,7 @@ public function getValue($columnName)
         // single-valued properties. We join the values when it's a string
         // for multi-values boolean/binary values we can't provide a
         // defined result so we return null
-        if (is_array($value)) {
+        if (is_array($value) && count($value)) {
             if (is_scalar($value[0]) && !is_bool($value[0])) {
```
dbu: the count should go as first check on this line, not in the outer if. we want to return null.

dbu:
there are no unit tests. we focussed on the functional phpcr-api-tests. those test a lot of this, though maybe not every edge case.

src/Jackalope/Query/Row.php
danrot: I am not sure if I understand you correctly. Should it be a separate if, or should I just swap the operands?

dbu: sorry. i meant to move count() into this if, so it becomes `if count($value) && is_scalar($value[0]) && !is_bool($value[0])) {`.

we should return null for an empty array.

danrot:
@dbu Is it now ok?

dbu:
looks good, thanks a lot!

danrot:
Thank you for merging :-)

PR #324, opened 2017-01-19
I tried to fix #314

### LINKED ISSUE #314: Invalid argument when throwing exception for unknown namespace:
The namespace collection can be an `ArrayObject`, in which case `array_keys` does not work, causing an error to be produced when an exception should be thrown in the `NodeProcessor`:

```
        if (!isset($this->namespaces[$alias])) {
            throw new NamespaceException(sprintf(
                'Namespace alias "%s" is not known for name "%s", known namespace aliases are: "%s"',
                $alias,
                $name,
                implode('", "', array_keys($this->namespaces))
            ));
        }
    }
```

vondrasekpavel:
I must update phpDoc in Jackalope-Doctrine-Dbal Client after merge :)

### DIFF
--- a/src/Jackalope/NodeType/NodeProcessor.php
+++ b/src/Jackalope/NodeType/NodeProcessor.php
@@ -342,21 +342,16 @@ $/xi";
      */
     private function validateNamespace($name)
     {
-        $alias = null;
         $aliasLength = strpos($name, ':');
 
-        if ($aliasLength === false) {
+        if (false === $aliasLength) {
             return;
-        } else {
-            $alias = substr($name, 0, $aliasLength);
         }
 
-        if (!isset($this->namespaces[$alias])) {
-            if ($this->namespaces instanceof ArrayObject) {
-                $this->namespaces = $this->namespaces->getArrayCopy();
-            }
+        $alias = substr($name, 0, $aliasLength);
 
-            $aliases = implode("', '", array_keys($this->namespaces));
+        if (!isset($this->namespaces[$alias])) {
+            $aliases = implode("', '", array_keys($this->namespaces->getArrayCopy()));
 
             throw new NamespaceException(
                 "Namespace alias '$alias' is not known for name '$name', known namespace aliases are: '$aliases'"
--- a/src/Jackalope/NodeType/NodeProcessor.php
+++ b/src/Jackalope/NodeType/NodeProcessor.php
@@ -2,6 +2,7 @@
 
 namespace Jackalope\NodeType;
 
+use ArrayObject;
 use PHPCR\ItemExistsException;
 use PHPCR\Lock\LockException;
 use PHPCR\NodeInterface;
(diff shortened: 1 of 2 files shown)


src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -343,19 +344,23 @@ private function validateNamespace($name)
     {
         $alias = null;
         $aliasLength = strpos($name, ':');
-        if (false === $aliasLength) {
+
+        if ($aliasLength === false) {
```
dbu: its a best practice to have the constant value on the left side and the variable right, to avoid accidentally assigning values. please revert this line

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -71,18 +72,18 @@ class NodeProcessor
...
-     * @param array   $namespaces       List of namespaces in the current session. Keys are prefix, values are URI.
-     * @param boolean $autoLastModified Whether the last modified property should be updated automatically
+     * @param string      $userId           ID of the connected user
+     * @param ArrayObject $namespaces       List of namespaces in the current session. Keys are prefix, values are URI.
+     * @param bool        $autoLastModified Whether the last modified property should be updated automatically
      */
     public function __construct(
         $userId,
-        $namespaces = array(),
+        ArrayObject $namespaces,
```
dbu: this makes $namespaces not optional anymore. but you verified usage of NodeProcessor and the namespaces are always passed in?

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -343,19 +344,23 @@ private function validateNamespace($name)
...
         $alias = null;
         $aliasLength = strpos($name, ':');
-        if (false === $aliasLength) {
+
+        if ($aliasLength === false) {
             return;
-        } 
-        
-        $alias = substr($name, 0, $aliasLength);
+        } else {
```
dbu: no need for an else. when the `if` is entered, we return. not having the else makes the code more readable as you see that we either already stopped, or we get to this line.

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -343,19 +344,23 @@ private function validateNamespace($name)
...
+        }
 
         if (!isset($this->namespaces[$alias])) {
-            throw new NamespaceException(sprintf(
-                'Namespace alias "%s" is not known for name "%s", known namespace aliases are: "%s"',
-                $alias,
-                $name,
-                implode('", "', array_keys($this->namespaces))
-            ));
+            if ($this->namespaces instanceof ArrayObject) {
```
dbu: can this ever not be true now? when we force ArrayObject and its not optional, i think we need no check.

src/Jackalope/NodeType/NodeProcessor.php
```diff
@@ -343,19 +344,23 @@ private function validateNamespace($name)
...
+                $this->namespaces = $this->namespaces->getArrayCopy();
```
dbu: this is dangerous, as it breaks the binding with the ArrayObject. after such an error, the behaviour will change.

i would do: `$aliases = implode("', '", array_keys($this->namespaces->getArrayCopy());`

src/Jackalope/NodeType/NodeProcessor.php
vondrasekpavel: OK :)

src/Jackalope/NodeType/NodeProcessor.php
vondrasekpavel: Sorry I forgot :(

src/Jackalope/NodeType/NodeProcessor.php
vondrasekpavel: Yes, I thought over this line... but I wanted to know your feedback :) I will update it :)

vondrasekpavel:
snimek obrazovky 2017-01-20 v 14 50 21

I tested it with Doctrine Dbal (I copied it to vendor) and all test worked.

src/Jackalope/NodeType/NodeProcessor.php
vondrasekpavel: I tried to search where is NodeProcessor used and I think that is actually only in Jackalope Doctrine Dbal Client and NodeProcessorTest:

snimek obrazovky 2017-01-20 v 14 54 46

And there aren't namespaces optional so I don't know...

src/Jackalope/NodeType/NodeProcessor.php
vondrasekpavel: Ok

src/Jackalope/NodeType/NodeProcessor.php
vondrasekpavel: Yes, you're right I will fix it :)

dbu:
great, thanks a lot!

vondrasekpavel:
I thank you too! :)

Issue #326, opened 2017-03-07
Broken ImportRepositoryContentTest


vondrasekpavel:
Hi,

I tested Doctrine Dbal (from dev-master ) against Jackalope dev-master and I found that is something wrong with this commit because tests say:

```
PHPCR\Tests\Import\ImportRepositoryContentTest::testImportXMLUuidReplaceRoot
RuntimeException: Object path for UUID "560fb205-d2e4-43e9-b0b3-5035a0b99edf" has already been registered to "/"
/Users/pavel/Sites/jackalope-doctrine-dbal/vendor/jackalope/jackalope/src/Jackalope/ObjectManager.php:1888
/Users/pavel/Sites/jackalope-doctrine-dbal/vendor/jackalope/jackalope/src/Jackalope/Node.php:625
/Users/pavel/Sites/jackalope-doctrine-dbal/vendor/jackalope/jackalope/src/Jackalope/ImportExport/ImportExport.php:293
/Users/pavel/Sites/jackalope-doctrine-dbal/vendor/jackalope/jackalope/src/Jackalope/ImportExport/ImportExport.php:581
/Users/pavel/Sites/jackalope-doctrine-dbal/vendor/jackalope/jackalope/src/Jackalope/ImportExport/ImportExport.php:124
/Users/pavel/Sites/jackalope-doctrine-dbal/vendor/jackalope/jackalope/src/Jackalope/Session.php:545
/Users/pavel/Sites/jackalope-doctrine-dbal/vendor/phpcr/phpcr-api-tests/tests/Import/ImportRepositoryContentTest.php:317
```

It throws RuntimeException from ObjectManager::registerUuid. But it shouldn't throw this exception because in importXML is set ImportUUIDBehaviorInterface::IMPORT_UUID_COLLISION_REPLACE_EXISTING option.

In this PR is bug: https://github.com/jackalope/jackalope/pull/307

ping @dantleech

dbu:
@dantleech any chance you can fix this soon? otherwise i will need to revert #307 as the bug breaks importing documents.

dantleech:
Will try and have a look tomorrow evening

PR #327, opened 2017-03-09
cleanup

vondrasekpavel:
php 5.4 short array syntax
:: clas names
php 5.6+ compatibility
removed deprecations from tests
updated travis to php 5.6+
added modern phpunit to dev dependencies

I tested it with jackalope-dbal and tests fails only on https://github.com/jackalope/jackalope/issues/326 so I think all tests work.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,14 +16,16 @@
     ],
     "minimum-stability": "dev",
     "require": {
-        "jackalope/jackalope-transport": "*",
-        "php": ">=5.3.3",
+        "php": "^5.6|^7.0",
         "ext-xml": "*",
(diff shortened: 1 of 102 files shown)


composer.json
```diff
@@ -16,15 +16,17 @@
     ],
     "minimum-stability": "dev",
     "require": {
-        "jackalope/jackalope-transport": "*",
```
dbu: this virtual package requirement should not have been removed. fixing in 35bd2ec12c4cfc76d3f4d0676c01fa4d4aee6621

vondrasekpavel: Ah, sorry, I pushed temporary change :(

PR #328, opened 2017-03-14
Handle root node replacement correctly

### LINKED ISSUE #326: Broken ImportRepositoryContentTest

dantleech:
Fixes #326 

- Allow forcing the registration of a UUID via. a flag.
- To allow the UUID_REPLACE_EXISTING flag to be used when importing
data.

--

- Is it valid to register a UUID with a different node path?

I wonder if it is better to "ask" the object manager first, if it has a UUID (optionally with a path), and then register it if required.

### DIFF
--- a/src/Jackalope/ImportExport/ImportExport.php
+++ b/src/Jackalope/ImportExport/ImportExport.php
@@ -292,8 +292,11 @@ class ImportExport implements ImportUUIDBehaviorInterface
             } elseif ('jcr:created' === $name || 'jcr:createdBy' === $name) {
                 // skip PROTECTED properties. TODO: get the names from node type instead of hardcode
             } elseif ('jcr:uuid' === $name) {
-                //avoid to throw an exception when trying to set a UUID when importing from XML
-                $node->setProperty($name, $info['values'], $info['type'], false);
+                // The root node is not removed and thus UUID needs not be updated
+                if (0 !== $node->getDepth()) {
+                    // Set the UUID on this node. Validate false to allow setting the UUID manually.
+                    $node->setProperty($name, $info['values'], $info['type'], false);
+                }
             } else {
                 $node->setProperty($name, $info['values'], $info['type']);
             }
--- a/src/Jackalope/Node.php
+++ b/src/Jackalope/Node.php
@@ -551,8 +551,8 @@ class Node extends Item implements IteratorAggregate, NodeInterface
     /**
      * {@inheritDoc}
      *
-     * @param boolean $validate does the NodeType control throw an exception if
-     *      the property can't be set? To use in case of UUID import
+     * @param boolean $validate When false, node types are not asked to validate
+     *                          whether operation is allowed
      *
      * @throws InvalidItemStateException
      * @throws NamespaceException
--- a/src/Jackalope/ObjectManager.php
+++ b/src/Jackalope/ObjectManager.php
@@ -1873,7 +1873,7 @@ class ObjectManager
      */
     public function registerUuid($uuid, $absPath)
     {
-        if (isset($this->objectsByUuid[$uuid])) {
+        if (array_key_exists($uuid, $this->objectsByUuid)) {
             throw new RuntimeException(sprintf(
                 'Object path for UUID "%s" has already been registered to "%s"',
                 $uuid, $this->objectsByUuid[$uuid]


dbu:
thanks. the current state of the PR would restore the previous behaviour
i think.

i just re-read
https://docs.adobe.com/content/docs/en/spec/jcr/2.0/11_Import.html

11.8.3 Replace Existing Node

IMPORT_UUID_COLLISION_REPLACE_EXISTING: If an incoming non-shareable
node has the same identifier as a node already existing in the
workspace, then the already existing node is replaced by the incoming
node in the same position as the existing node. Note that this may
result in the incoming subgraph being disaggregated and “spread around”
to different locations in the workspace. In the most extreme case this
behavior may result in no node at all being added as child of
parentAbsPath. This will occur if the topmost element of the incoming
XML has the same identifier as an existing node elsewhere in the
workspace. In the case of shareable node, however, the behavior differs
(see §14.1.2 Shared Node Creation on Import).

i think we can ignore shareable node. i interpret the rest of the doc as
saying if i add a node /foo/bar with uuid $a, but $a already exists at
/other/path, the new node would replace /other/path and bring its
children there, rather than being added where we asked for it and "move"
the uuid to /foo/bar. so indeed the importing code would need to check
for existing uuid and handle this as a special case. the spec
unfortunately does not mention what happens with the children of the
existing node. i would assume we merge things, but as we don't support
same-name siblings this could lead to collisions on that. which might
break existing code, but i don't see a way to avoid that - if you
exported, then manipulated the data and relied on this bug then its
tough luck. we need to explain it in the changelog for 1.3 however.

does that make sense to you, do you agree? could you see working on
that? its of course more than what things looked like when you started
registering the uuid... its technically a different bug, so i would be
happy to at least merge the PR you provide here, as it fixes the
regression and will make the acceptance tests work again.

dantleech:
Yeah, I think this PR restores original behaviour, so happy if you're happy :)

On 14 March 2017 21:05:25 GMT+00:00, David Buchmann <notifications@github.com> wrote:
>thanks. the current state of the PR would restore the previous
>behaviour
>i think.
>
>i just re-read
>https://docs.adobe.com/content/docs/en/spec/jcr/2.0/11_Import.html
>
>11.8.3 Replace Existing Node
>
>IMPORT_UUID_COLLISION_REPLACE_EXISTING: If an incoming non-shareable
>node has the same identifier as a node already existing in the
>workspace, then the already existing node is replaced by the incoming
>node in the same position as the existing node. Note that this may
>result in the incoming subgraph being disaggregated and “spread around”
>to different locations in the workspace. In the most extreme case this
>behavior may result in no node at all being added as child of
>parentAbsPath. This will occur if the topmost element of the incoming
>XML has the same identifier as an existing node elsewhere in the
>workspace. In the case of shareable node, however, the behavior differs
>(see §14.1.2 Shared Node Creation on Import).
>
>i think we can ignore shareable node. i interpret the rest of the doc
>as
>saying if i add a node /foo/bar with uuid $a, but $a already exists at
>/other/path, the new node would replace /other/path and bring its
>children there, rather than being added where we asked for it and
>"move"
>the uuid to /foo/bar. so indeed the importing code would need to check
>for existing uuid and handle this as a special case. the spec
>unfortunately does not mention what happens with the children of the
>existing node. i would assume we merge things, but as we don't support
>same-name siblings this could lead to collisions on that. which might
>break existing code, but i don't see a way to avoid that - if you
>exported, then manipulated the data and relied on this bug then its
>tough luck. we need to explain it in the changelog for 1.3 however.
>
>does that make sense to you, do you agree? could you see working on
>that? its of course more than what things looked like when you started
>registering the uuid... its technically a different bug, so i would be
>happy to at least merge the PR you provide here, as it fixes the
>regression and will make the acceptance tests work again.
>
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/jackalope/jackalope/pull/328#issuecomment-286560260

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

dbu:
i realized that we already handle IMPORT_UUID_COLLISION_REPLACE_EXISTING correctly. this was just a super edge case when the import replaces the root node. as the root can't be deleted, we instead remove all its properties and children and then import onto that. but because the root node is not deleted, the uuid stays. i changed the code to not try to re-set the uuid of the root node as thats not needed.

PR #329, opened 2017-03-23
Fixed fallback-definition to return important defintion first

wachterjohannes:
When receiving `NodeDefinitions` from a node which has the `nt:unstructured` as primary-type. The resulting definition is always the `nt:unstructured` definition.

Also if the node would match multiple definitions in the chain (because of matching the wildcard).

The comment in the line 541 describes also my problem. But my definitions defer in the "OnParentVersionAction" which is quite important for our usecase in sulu.

### DIFF
--- a/src/Jackalope/Item.php
+++ b/src/Jackalope/Item.php
@@ -579,7 +579,7 @@ abstract class Item implements ItemInterface
                 if ($candidate->getName() === $this->name) {
                     return $candidate;
                 }
-                if ('*' === $candidate->getName()) {
+                if (!$fallbackDefinition && '*' === $candidate->getName()) {
                     // if we have multiple wildcard definitions, they are hopefully equivalent
                     $fallbackDefinition = $candidate;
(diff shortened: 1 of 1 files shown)


dbu:
hm, maybe i was too quick to merge. are you sure this fixes anything for you? as we return immediately when we find a definition that has the same name for the property, we return it immediately, so wildcard should never overwrite a specific definition. the only difference your change seems to make is that we pick up the first `*` definition we see instead of the last. what am i missing?

dbu:
i will keep this, as at the very least its a small performance improvement to check for null instead of string comparison.

wachterjohannes:
@dbu thanks for merging.

dbu:
thanks, so you want the most specific definition. that makes sense when
you need more than just "any" definition that tells it allows the field.

wachterjohannes:
yes our own definition has another value for `OnParentVersionAction` as the `nt:unstructured` definition.

PR #333, opened 2017-04-10
get string representation of value for error message without exceeding memory limit

danrot:
The `var_export` method did exceed the memory limit in our tests. This should fix it.

### DIFF
--- a/src/Jackalope/NodeType/NodeType.php
+++ b/src/Jackalope/NodeType/NodeType.php
@@ -215,7 +215,7 @@ class NodeType extends NodeTypeDefinition implements NodeTypeInterface
                     }
 
                     if (is_array($value)) {
-                        throw new ConstraintViolationException('The value '.var_export($value, true) .' is multivalued, but the property definition for ['.$this->getName()."]:$propertyName is not.");
+                        throw new ConstraintViolationException('The value '.$this->getValueAsString($value).' is multivalued, but the property definition for ['.$this->getName()."]:$propertyName is not.");
                     }
                 }
 
@@ -263,7 +263,8 @@ class NodeType extends NodeTypeDefinition implements NodeTypeInterface
             }
         }
         if ($throw) {
-            $val = is_object($value) ? get_class($value) : (is_scalar($value) ? (string) $value : is_array($value) ? var_export($value, true) : gettype($value));
+            $val = $this->getValueAsString($value);
+
             throw new ConstraintViolationException("Node type definition ".$this->getName()." does not allow to set the property with name '$propertyName' and value '$val'");
         }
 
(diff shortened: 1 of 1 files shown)


src/Jackalope/NodeType/NodeType.php
```diff
@@ -392,4 +393,28 @@ public function canRemoveProperty($propertyName, $throw = false)
...
+     */
+    private function getValueAsString($value)
+    {
+        if (is_object($value)) {
+            $val = get_class($value);
+        } elseif (is_scalar($value)) {
+            $val = (string)$value;
+        } elseif (is_array($value)) {
+            $val = [];
+            foreach ($value as $k => $v) {
```
danrot: @dbu Was this the change you've been talking about?

src/Jackalope/NodeType/NodeType.php
```diff
@@ -392,4 +393,28 @@ public function canRemoveProperty($propertyName, $throw = false)
 
         return true;
     }
+
+    /**
+     * @param $value
```
dbu: `mixed`

please also add a short description to the method, like "Get a string representation of a value for error reporting."

src/Jackalope/NodeType/NodeType.php
```diff
@@ -392,4 +393,28 @@ public function canRemoveProperty($propertyName, $throw = false)
...
+     * @return array|mixed|string
+     */
+    private function getValueAsString($value)
+    {
+        if (is_object($value)) {
+            $val = get_class($value);
```
dbu: can you please rewrite this to return early? this line would become `return get_class($value);`.

and we can get rid of the elseif below and just do if again.

src/Jackalope/NodeType/NodeType.php
```diff
@@ -392,4 +393,28 @@ public function canRemoveProperty($propertyName, $throw = false)
...
+        } elseif (is_scalar($value)) {
+            $val = (string)$value;
+        } elseif (is_array($value)) {
+            $val = [];
+            foreach ($value as $k => $v) {
+                if (is_object($v)) {
+                    $val[$k] = get_class($v);
```
dbu: this will omit all values that are not objects. i think we should simplify this to `return 'array of length '.count($value);` instead of trying to show whats inside the array.

danrot:
@dbu Adapted according to your comments. If I got you right there is no `var_export` at all anymore, hope that is correct this way.

src/Jackalope/NodeType/NodeType.php
```diff
@@ -392,4 +393,28 @@ public function canRemoveProperty($propertyName, $throw = false)
 
         return true;
     }
+
+    /**
+     * Get a string representation of the passed value for error reporting.
+     *
+     * @param mixed $value
+     *
+     * @return array|mixed|string
```
dbu: this method always returns a string.

danrot:
@dbu Done :-) It would be great if you could release that as a hotfix soonish 😃

dbu:
thank you, i will tag another patch version now. 1.3 seems not be a lucky release, i think this is the 3rd patch release since last week :-/

danrot:
@dbu That's also my fault, could already have fixed that immediately after 1.2.8 was release 😕

wachterjohannes:
@dbu is it possible to port that to 1.2 release series?

dbu:
sure. cherry-picked and tagged 1.2.9

dbu:
(btw if you think of that before doing the pull request, you can pull requests against the 1.2 branch instead of master)

wachterjohannes:
@dbu for we thought about updating to 1.3 in sulu anyway. but we saw too late that you dropped 5.5 support.

Thanks for fast response (:

PR #337, opened 2017-06-22
Fix typo in DebugStack

relo-san:
This typo summons Doctrine annotation reader for search nonexistent "retrun" annotation, which produce error:
[Semantical Error] The annotation "@retrun" in method Jackalope\Transport\Logging\DebugStack::getBacktrace() was never imported. Did you maybe forget to add a "use" statement for this annotation?


dbu:
darn, i am usually not legastenic, but retrun is something i seem to love to do. thanks for the fix!

Issue #346, opened 2018-02-16
getVersionByLabel does not work due to SystemItem check?


UFTimmy:
I have made a checkpoint for a document, and given it a label like so:

        $dm->checkpoint($node);

        $dm->getPhpcrSession()
            ->getWorkspace()
            ->getVersionManager()
            ->getVersionHistory($node->getId())
            ->addVersionLabel('1.0', 'live', true);

That works, but I cannot find the version later. I am using this to find the version:

        $versionHistory = $this->getDocumentManager()->getPhpcrSession()
            ->getWorkspace()
            ->getVersionManager()
            ->getVersionHistory($contentDocument->getId());

        $page = $versionHistory->getVersionByLabel('live');

When I step into the code, I believe the problem lies with the check to see if the Node is a system node here: https://github.com/jackalope/jackalope/blob/master/src/Jackalope/Version/VersionHistory.php#L283

You will have to forgive me, because I am not terribly familiar with Jackrabbit/Jackalope and how nodes and properties work, but this is what I see using Xdebug.

In this method, $node->getProperties has two properties:
jcr:primaryType (nt:versionLabels) and a property called live, which appears to be the label I created, it has a correct reference to the version.

So it appears adding the version label is working. And if I alter the check to see if it's a system node, everything works.

Like I said, I am not familiar with nodes in general, but the first thing isSystemItem does it check the depth, and if it's greater than 1, it's not a system item. This fails for both properties, as they have depths of 8. You can see that here: https://github.com/phpcr/phpcr-utils/blob/master/src/PHPCR/Util/NodeHelper.php#L131

Is the call to isSystemItem incorrect? It seems odd to me that it's checking if the node itself is a system item inside a loop of the node's properties, instead of doing that check just once outside the loop. But changing it to check if the property is a system item does not work.

Basically, I believe I have found a problem, but I am not sure of the solution, though I would be happy to help if someone could point me in the right direction.

dbu:
thanks for the detailed bug report. your analysis is great, i think you are absolutely right.

i think i found the place where i told the author to use that method, not realizing that its wrong... https://github.com/jackalope/jackalope/pull/302#discussion_r45623324

i guess our test is using a node on top level, and maybe at the time this was contributed, the helper method did not check for the depth. can you replace that line by `if ('jcr:primartyType' !== $property->getName()) {` and do a pull request with that if it fixes the problem?

UFTimmy:
That fixed it, so I created the PR. Thank you!

PR #347, opened 2018-02-19
Fix for initVersionLabels

UFTimmy:
Can't use isSystemItem because it checks the depth.

See #346


dbu:
great, thanks a lot! i will fix the styleci error and then tag 1.3.5 with this fix

dbu:
there we go: https://github.com/jackalope/jackalope/releases/tag/1.3.5

PR #356, opened 2020-02-17
Fix type error on php7.4

alexander-schranz:
In some cases it seems that a node path is a integer.

### DIFF
--- a/tests/Jackalope/ObjectManagerTest.php
+++ b/tests/Jackalope/ObjectManagerTest.php
@@ -57,10 +57,10 @@ class ObjectManagerTest extends TestCase
 
     public function testRegisterUuidAlreadyMapped()
     {
+        $this->om->registerUuid('1234', '/path/to/this');
+
         $this->expectException(\RuntimeException::class);
         $this->expectExceptionMessage('Object path for UUID "1234" has already been registered to "/path/to/this"');
-
-        $this->om->registerUuid('1234', '/path/to/this');
         $this->om->registerUuid('1234', '/path/to/that');
     }
 }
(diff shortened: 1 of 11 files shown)


dbu:
i guess the stupid behaviour of php to convert array keys to int when they look like int...

the fix seems right to me. to make the tests green on php 7.4, we would need to upgrade phpunit :-| 
can you please try to do that? i think we should make this a new minor version and drop php 5.6/7.0 support to not overcomplicate this.

please add notes to the CHANGELOG.md file too.

alexander-schranz:
@dbu not sure what I did wrong that travis is not longer trigger 🙈 maybe it wants to keep php 5.5 and 7.0 😄

tests/Jackalope/ObjectManagerTest.php
```diff
@@ -55,12 +55,11 @@ public function testRegisterUuid()
...
     }
 
-    /**
-     * @expectedException \RuntimeException
-     * @expectedExceptionMessage Object path for UUID "1234" has already been registered to "/path/to/this"
-     */
     public function testRegisterUuidAlreadyMapped()
     {
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage('Object path for UUID "1234" has already been registered to "/path/to/this"');
```
dbu: please put this after the first `registerUuid` call. the upside of having the expect exceptions in code is that they can be at the precisely correct place.

alexander-schranz: done

dbu:
> could you please add something to the changelog file?

the changelog is still missing, otherwise looks good to merge to me.

alexander-schranz:
@dbu ups forget to push is now there.

dbu:
thanks!

dbu:
and tagged https://github.com/jackalope/jackalope/releases/tag/1.4.0

PR #357, opened 2020-02-26
Fixed typo in ObjectManager PHPDoc

muffe:
This erroneous pipe is currently crashing psalm in my setup.


dbu:
ups, weird one :confused: thanks for the fix.

PR #369, opened 2022-04-20
Use symfony 6 phpunit bridge

alexander-schranz:


dbu:
thanks. do you need a release with this?

PR #381, opened 2023-11-16
Added return annotations to many methods that implement PHPCR interfaces

stollr:
The Symfony deprecation helper reports several deprecation warnings telling that future versions of the PHPCR interfaces may add native return values in the future. This is fixed by the added return annotations.

I'd be happy if this could be merged soon.

### DIFF
--- a/src/Jackalope/Item.php
+++ b/src/Jackalope/Item.php
@@ -218,6 +218,8 @@ abstract class Item implements ItemInterface
     /**
      * {@inheritDoc}
      *
+     * @return string the normalized absolute path of this Item
+     *
      * @api
      */
     public function getPath()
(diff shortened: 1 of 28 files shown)


dbu:
hm, is `@inheritdoc` not working? the interface defines the `@return`.

in version 2 we will have the strict type declarations, e.g.: https://github.com/jackalope/jackalope/blob/8dce45068df57bf283760fd3f3d840c0f465ac3a/src/Jackalope/Item.php#L194

but i can merge this here if it is needed to avoid hickups. its redundant to the doc from the phpcr interfaces. are you sure its needed?

stollr:
Symfony's deprecation helper does not respect `@inheritdoc` by design. You can read more about it [here](https://github.com/symfony/symfony/issues/44802).

So it would be nice if you could merge it :-)

dbu:
here you go https://github.com/jackalope/jackalope/releases/tag/1.4.7

for 2.0 i revert the change again as we add the actual return type declarations https://github.com/jackalope/jackalope/pull/382

stollr:
Thank you 👍

